PR #1, opened 2021-03-10
Proper Dark Mode Support

Oshgnacknak:
These four CSS lines are enough for some nice Dark Mode action.
image
- Osh

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -1,4 +1,6 @@
 body {
+        background-color: #151515 ;
+        color: white ;
 	max-width: 800px ;
 	margin: auto ;
 	margin-bottom: 500px ;
@@ -16,4 +18,12 @@ img {
 	max-width: 600px ;
 	margin: auto ;
 	display: block ;
+}
+
+a {
+        color: lightblue ;
+}
+
+a:visited {
+        color: gray ;
 }


LukeSmithxyz:
Okay. I'm going to leave this open for a bit to think about it or see if others have opinions.

I am 100% pro-dark (and think it should always be default), but I don't know if it's a turn off for some people.

DominikRusso:
Using [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) might be an alternative.

LukeSmithxyz:
How commonly is that actually set? Is that typical in browsers or something without me knowing?

LukeSmithxyz:
Oh, actually, it looks like I already have dark mode as default without even remembering. I think I'll go ahead and add these settings in a `prefers-color-scheme` statement. I am still open to the idea of dark as default though.

Also, I'm still open to whatever kind of tweaks to either the dark or light color schemes.

DominikRusso:
In normie operating systems (win and mac) the browser gets the preference from the OS. In linux you can enable it by passing `--force-dark-mode` or enabling it in the flags (`chrome://flags/#enable-force-dark`).

0neGal:
> How commonly is that actually set? Is that typical in browsers or something without me knowing?

More than you'd think, iOS 13 and up sets it, Windows and macOS has been setting it since forever, and your QT and GTK theme also sets it automatically in most browsers.

The preferable way to handle this would be with:
```css
@media (prefers-color-scheme:dark) {
    body {
        color: white;
        background-color: black;
    }
}
```

Obviously adjusting the colors to your liking.

You can also furthermore add:
```css
@media (prefers-color-scheme:light) {
    body {
        color: black;
        background-color: white;
    }
}
```
To fix issues like this
2021-03-10_Wed-Mar_09-29-34

In my case QuteBrowser can only set the default browser background and not foreground, so some sites look like this.

PR #4, opened 2021-03-10
Add oats

m-chrzan:
Adds oats.


LukeSmithxyz:
Idk, oats are p cringe/non-based. I'll think about it.

LukeSmithxyz:
The only reason I included a pasta page is because it is an ingredient in other dishes that a NEET somehow might not know how to boil. I would rather link the page from other pages rather than routinely recite the same directions.

twiclo:
Oats are great for bulking which is when you should be eating carbs/pasta

LukeSmithxyz:
I'll give in, but I stand by that oat consumption is always cringe.

PR #5, opened 2021-03-10
Added carbonade flamande (very nice traditional beef stew)

ptethng:


ptethng:
These are not cloves of garlic, they're just called cloves. They are basically dried flower buds that are commonly used as a spice or as a cure for a toothache.

I'm going to work on those American measurements, thanks for the feedback.

LukeSmithxyz:
Oh okay, I just looked it up. I've never heard of that, but I guess it's a thing.

ptethng:
There you go, amended my commit.

Yeah, maybe they're a bit uncommon in America. They kinda taste like a mix of very sharp cinnamon without the sweetness, and they really work quite well for toothaches, although they kinda taste horrible when you just bite into them

PR #6, opened 2021-03-10
Adding Portuguese Steak with Beer Sauce recipe

lisbonjoker:


LukeSmithxyz:
Could you remove salt and pepper as ingredients?

I haven't been including them because it'd just be two lines in literally every page. I'd like that to be the standard.

lisbonjoker:
Yes, will do!

PR #11, opened 2021-03-10
Tuscan Style Pork Roast and typo fix

twiclo:
I've always wanted recipes to have the format of "now use ingredient(amount so you don't have to look up to the top of the page)" so I'm trying that out here. I also think recipes need to have a prep time and serving size


LukeSmithxyz:
who's => whose

LukeSmithxyz:
Well I should also ask is "nutorious" a Babish inside joke or a mispelling?

twiclo:
I wish I was that clever

PR #25, opened 2021-03-10
Remove page margin when printing

Wasby:
I personally prefer this when printed on paper.

Examples:
before: [https://0x0.st/-ZBB.pdf](https://0x0.st/-ZBB.pdf)
after: [https://0x0.st/-ZBu.pdf](https://0x0.st/-ZBu.pdf)

### DIFF
--- a/src/style.css
+++ b/src/style.css
@@ -33,3 +33,9 @@ img {
 		color: tomato ;
 	}
 }
+
+@media print {
+	body {
+		max-width: none ;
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Usually I would prefer that too, especially for pages full of text, but since most of the content here is mere lists rather than paragraphs, I think the margins help to center it a little better and are more sightly print-outs.

PR #39, opened 2021-03-11
MealDB Recipes

Mr-X-666:
I have made all the markdown recipes from openrecipes.xyz to the markdown format used here and added them to index.md


LukeSmithxyz:
I'll look at this later, but for one, please only use proper filenames.
No capitals, no spaces (use hyphens like the others), no ampersands or other symbols.

I can also see that many of these are not fully converted. Some have item numbers on separate lines than the directions.

LukeSmithxyz:
I suppose it should also be decided what to do with the mainpage with all these addtions.

Mr-X-666:
Yeah, the mainpage will be very hard to read through with these many recipes.
I'll change the filenames and fix those recipes rn.

m-chrzan:
Two comments on this:

1. Depending on licensing, would be good to copy images to here so as to not rely on themealdb.com staying up.
2. We probably can't just "public domain" these. Not a lawyer, but I imagine it would be good to add something like "everything on this site is public domain unless otherwise specified" to the footer blurb, and then add some short mention of this coming from TheMealDB. Unfortunately couldn't find any mention of licensing on their website.

Mr-X-666:
Ok I'll add the mention to mealDB in mealDB recipes. On the topic of images, I think we'll to wait till we decide the final categorized  file tree for images and recipes in general.
I''ve fixed the recipes with anomalies by going through each and every one of them and have minimalized file names

Mr-X-666:
Btw, I have arranged recipes in index.md in alphabetical order, including the recent recipe merges

Mr-X-666:
Closing PR due to huge influx of recipes

PR #56, opened 2021-03-11
Add banana bread

m-chrzan:


russellpierce:
@m-chrzan the README.md says:

> If measurements are used, include metric and American measurements. Do not use decimals in American; round or use fractions.

So, I think your PR might benefit from metric conversions on the ingredients for this PR, e.g. https://github.com/LukeSmithxyz/based.cooking/blob/ea2b2e66bb81dbeefb091be0778de42593a14284/src/french-mustard-sauce-porkchops.md

m-chrzan:
Anyone with a metric scale and graduated cylinder should also have a teaspoon, tablespoon, and a cup (or glass or mug) in their kitchen (;

I think it's necessary to include conversions for formal units (e.g. oz vs. ml, lbs vs. kg), but this recipe has informal measurements that are system-agnostic.

lucasfeijo:
I was gonna post a similar recipe, mine differs in that it has only 1.5 cup flour, and 0.5 cup sour cream. Also, no spices, just vanilla.

PR #71, opened 2021-03-11
Add Älplermagronen

AlexBocken:
Wanted to contribute with one of my favorite recipes.

`index.md` seems to be just chronological for now, right? I just appended mine at the bottom.


AlexBocken:
Okay, I've finally turned on spellcheck in vim. Clearly, I need it.

PR #78, opened 2021-03-11
Makefile-based generation system

sylGauthier:
Use a [friend's pretty cool blog Makefile system](https://pedantic.software/git/blogit) to handle updates better (don't regenerate already existing pages), tag system for easy navigation, RSS feeds, etc, all self contained in a less than 200 lines Makefile.

See commit message on  171d5a7 for more details.
See here the result: [https://sgauthier.fr/cooking/](https://sgauthier.fr/cooking/)

Also, tagged all existing recipes.

Note: will require some edit if you do want to go through with the merge, especially the config file where you need to specify the remote (where the site is actually hosted) for rsync, and templates to replace the fillers.

Note 2: I removed the tips pages because they don't really fit in the existing system, I can re-add them later by extending the build system.

Note 3: for the search system, I will generate a plain file with `<URL> <name of the recipe>` on each line for all recipes, then pipe it into my [choice](https://pedantic.software/projects/choice.html) tool to get a cool search function.

syg (at) pedantic.software

Cheers!


LukeSmithxyz:
Okay, very nice. This is an interesting implementation.

I really don't like there being two titles. I don't really want the name of the site at the top of every page. The link to the index and the metadata should be either under the recipe title or at the bottom. I'm sure there will be some injection required to get them under the title.

Note also the page names: they have `# ` in them, so that should be extracted and I think the website name can be added after the page name and a en dash like it is now.

sylGauthier:
All done I think, you can check my link again to see if that's better.

By the way, to build it requires `markdown` in your path, I recommend [this one](https://www.pell.portland.or.us/~orc/Code/discount/) because it's in a real language (C) and not some perl abomination. It's in gentoo's official repo (`app-text/discount`) so I would expect it to be available for most distros.

You'll probably want to edit `config` and add:
```
BLOG_REMOTE:=root@based.cooking:/var/www/cooking
```

LukeSmithxyz:
So you say it doesn't regenerate existing pages. Does this mean it will not update if a small edit is made to a page?

LukeSmithxyz:
This might be a little much, but would it be easily possible to generate a tag list of links at the bottom of each recipe page that links to the tag pages?

sylGauthier:
> So you say it doesn't regenerate existing pages. Does this mean it will not update if a small edit is made to a page?
It's really just working like a Makefile. So yes, if a markdown page is newer (as in, "modified", not "created") than its html output, it will be regenerated. And that's just what you get for free for using a Makefile. And if a HTML template used in a page is modified (such as the header, footer or what not), the page will also be regenerated.

> This might be a little much, but would it be easily possible to generate a tag list of links at the bottom of each recipe page that links to the tag pages?
That's probably easy. I'll give it a go.

sylGauthier:
I added the tag links at the end of each recipe, rebased on all new articles, tagged all new articles (plus fixed some typos, people really need to turn on syntax highlighting for markdown files on vim, maybe you can make a video on that :P).

I amended `README.md` and `example.md` to fit the new system.

It's theoretically mergeable as-is and should work on your side out-of-the-box.

LukeSmithxyz:
Okay, great. So one last detail. I've changed the way the repo is deployed, so now Github automatically logs into my server and runs the commands at the bottom of `.github/workflows/upload.yml` when this thing is updated. Could you replace those commands with whatever is needed to update with this version? It seems like the deploy command still uses rsync.

sylGauthier:
Sure! I can keep the current idea of ssh-ing into based.cooking and generating everything there, just a few things tho:
- I can't figure out from reading the `upload.yml` file what's the deployment full path (from what folder the HTML files are actually served). I would need this for the `config` file (the `rsync` will then be done locally, which is fine).
- You'll need `discount` as mentioned earlier, it should be in ubuntu repos AFAICT.
- I can specify a `-jN` option to the make command to make it run on multiple parallel cores so let me know if you think that's useful, and how many cores can be used on your server.

LukeSmithxyz:
- I can just change what directory is served by the server. I guess we can say `/var/www/based.cooking` should be it and I will adjust it accordingly.
- Already done.
- I don't think that would be necessary in this case. While it's on multiple cores now, I might move it to a place where it only has one. I don't think it would make a huge difference either way considering.

sylGauthier:
Rebased, added `index.md` into the index generation process (gets appended after the recipe list).

Updated footer.

Updated `upload.yml` and `config` to make it work on based.cooking.

LukeSmithxyz:
Okay. I'm going to go ahead and merge this and check for issues. It won't be live on the site yet, but it'll give me time to play around.

Thanks hugely. This seems like a great tool that will make things easier in the long run.

LukeSmithxyz:
It's live on the site now.

Looks generally functional.

I noticed, however, that the `article_list_footer.html` isn't appearing in the final index...

sylGauthier:
Whoops, did you just do a `make init`?

Don't do that it'll overwrite all templates.

sylGauthier:
> I noticed, however, that the `article_list_footer.html` isn't appearing in the final index...

#162 fixes this. My mistake.

LukeSmithxyz:
> Whoops, did you just do a `make init`?
> 
> Don't do that it'll overwrite all templates.

Lol yeah I noticed.

LukeSmithxyz:
All fixed now.

I can probably do this myself tomorrow, but I'll go ahead and bring it up to you since you just put this together:

There are two obvious things with the RSS/atom feeds. Firstly, they include the `# ` in the entry name. Secondl, frankly I'd like the entire content of the page in the RSS entry. I don't think it'd be too much and I know I prefer reading things in the RSS feeder rather than clicking on a site. They would have to be in HTML too to be formated correctly. Right now it's just taking the stuff above the ingredients section which is noting or an unformatable markdown image or link.

PR #86, opened 2021-03-12
add the first draft of the basic sweet potato fries recipe

shanekunz:


shanekunz:
I added a personal webp image of the finished recipe to the "pix" folder, but I did not reference the photo as I don't see an example of that.

m-chrzan:
See [here](https://github.com/LukeSmithxyz/based.cooking/blob/master/src/cacio-e-pepe.md) for image example. Basically

    <alt text>

Remove newlines between bullets under "Notes".

shanekunz:
cool, thanks. image ref added, newline removed.

PR #88, opened 2021-03-12
Add a recipe for banana muffins with chocolate.

MasFlam:
I hope it's not too "hyper-sugary." Those are good muffins.


Cloud11665:
Based recipie from a based man

PR #96, opened 2021-03-12
Disable deployments on PR creation

m-chrzan:
I'm pretty sure the `on: pull_request` event means deployments would be attempted from each PR, even before they're merged into master. Just having `on: push` should capture both direct pushes and PR merges to master. [Reference](https://stackoverflow.com/a/61565445/7406283).


LukeSmithxyz:
Okay. Either way it will stop all these annoying error messages I'm getting until I fix the thing.

PR #116, opened 2021-03-12
A bunch of recipes from my personal site

Yaroslav-95:
* Beef Goulash
* Cheesy Meatballs with Tomato Sauce
* Country Breakfast Skillet
* Creamy Mashed Potatoes
* Fresh Guacamole
* Merchant's Buckwheat
* Pasta Navy Style
* Pan-seared Chicken Fillet
* Parmesan Potatoes
* Refried Beans
* Sunday Morning Milkshake

Also accidentally modified the template with a find ... -exec sed
command, but I'll commit it since you mentioned that you accept webp's
only anyway ¯\_(ツ)_/¯.

Also I have a question for you Luke: Have you thought how to deal with the situation when someone wants to add a recipe that technically is already on the site? I ask because I have my own versions of Carbonara and Chilli con carne, but I haven't added them yet because I saw them on the site already.


LukeSmithxyz:
I think that we should stray from doubling up recipes, although if you feel that an already existing recipe could be improved, you can always add what you'd like.

Yaroslav-95:
Well, I wouldn't say that my version of, for example, Chili con carne is better or worse than the one currently on the site, it's just different. I do have a picture for it, but it's obviously of my version of the recipe, so I am not sure that it would be representative of the one currently on the site. I agree that there shouldn't be doubling up of recipes, but I also don't think it is completely fair to change the current recipe, there will be people that prefer this one to mine, for example. 

I was thinking, maybe a link to "alternative" versions could be provided inside the first such recipe that was submited to the site. Just an idea, maybe for the future.

PR #122, opened 2021-03-13
Oaty pancakes

puremana:


TomBT:
I think it'd be best to add Oaty Pancakes under the already created Pancake index like how Mexican Rice under the [Rice index](https://based.cooking/rice) and the Cheesey Omelet is under the [Omelet index](https://based.cooking/omelet).

puremana:
I've moved it to the pancake index but it's a little bit different because it's not just an extra thing, it's a completely different recipe and instructions

TomBT:
Ah I see

LukeSmithxyz:
This should really be a different article.

puremana:
Changed back to being a unique recipe...

PR #153, opened 2021-03-14
curl/Search function

Goblin80:
Implementing the curl/search functionality using a script that iterates over each word in the recipe's title and create a new file with the word as a the file name and the recipe link as a it's content:

```bash
#!/bin/bash

URL="https://based.cooking"

# generate search indices next to 'generate_search_index.sh'
search_dir=$(dirname "$0")

# before regenerating search indices
# clean search directory except 'generate_search_index.sh'
find $search_dir ! -name `basename $0` -type f -delete

# list all recipes in src/
for filepath in $search_dir/../*.md; do
    # grab filename of the recipe
    filename=$(basename $filepath)

    # grab filename of the recipe WITHOUT it's extension: '.md'
    title=${filename%.*}

    # split title into words delimited by dash '-'
    title_delimited=$(echo $title | tr '-' "\n")

    # iterate over each word in the title
    for word in $title_delimited; do
        # for each word, create a new file that contains links to it's recipes
        echo "$URL/$title.html" >> $search_dir/$word
    done
done

```

```bash
$ tree
.
├── example.md
├── README.md
├── src
│   ├── search
│   │   ├── aelplermagronen
│   │   ├── almeirim
│   │   ├── ...
│   │   ├── fries
│   │   ├── generate_search_index.sh
│   │   ├── ginataang
│   │   ├── ginger
│   │   ├── gnocchi
│   │   ├── ...
│   │   ├── yam
│   │   └── yogurt
```

after running `./src/search/generate_search_index.sh` before deployment, calling the endpoint `/search` with a keyword will return:

```bash
$ curl based.cooking/search/chicken

https://based.cooking/chicken-biscuit-potpie.html
https://based.cooking/chicken-parmesan.html
https://based.cooking/chicken-pasta-casserole.html
https://based.cooking/chicken-stock-bone-broth.html
https://based.cooking/chicken-tacos.html
https://based.cooking/chicken-tomato-spinach-curry.html
https://based.cooking/chicken-wings.html
https://based.cooking/pan-seared-chicken.html
https://based.cooking/roasted-chicken-breast.html
```

---

~~Also this PR has an attempt to fix the github's failing upload workflow by installing necessary packages first. but of course i cannot test that myself.~~ i am not sure if you are using github's workflow to upload to the live site..

Thanks @sylGauthier for the cool Makefile...


Godje:
Summary for reviewer.

  one recipe filename renamed to snake case.
  Readme.md changed.
  modification to the github actions script made that now updates and installs `wget`.
  actual script file added.

Godje:
Personally, I think this is creating too many files. I understand the approach of having it generated on deploy time. But I don't think the existence of so many files is helpful.
This whole thing on its own has a functionality of a whole separate package.

Goblin80:
PR has been adapted to the new makefile that @sylGauthier added.

Goblin80:
curl/Search now support looking up tags (same endpoint).

```bash
$ curl http://based.cooking/search/german

https://based.cooking/sauerkraut.html
```

LukeSmithxyz:
I haven't had a chance to actually test this out.

I'll go ahead and say, since we're talking about this being run from the command line, I'd prefer the `/search/` url be something shorter, like `/q/` or if possible, nothing at all and on 404, a search is run.

Additionally, I think it'd be nice to explicitly recommend people to make an alias or really function for for `curl based.cooking/q/%s` because I doubt people will want to type that out every time.

Again, I'll have to test if it actually works later.

Goblin80:
> I'll go ahead and say, since we're talking about this being run from the command line, I'd prefer the `/search/` url be something shorter, like `/q/` or if possible, nothing at all and on 404, a search is run.

There are some benefits to have a dedicated endpoint for searching:

1. Makes it easy to flush search index: simply `rm -rf search`
1. Avoid polluting blog directory.
1. Avoid any confusion that may occur if someone is looking for a **recipe with a single word** eg: *aelplermagronen*.

## Despite all of that...

I went ahead and **removed `/search/`**.

```bash
$ curl http://based.cooking/german

https://based.cooking/sauerkraut.html
```

> Additionally, I think it'd be nice to explicitly recommend people to make an alias or really function for for `curl based.cooking/q/%s` because I doubt people will want to type that out every time.

maybe something like this?

```bash
$ basedcooking () { curl -sL https://based.cooking/"$@" | fzf --preview 'curl -sL {} | bat --color=always' ; }
$ basedcooking foo
```

based-cooking-fzf-preview

<sub><sub>* curl + fzf + bat = bloat?</sub></sub>

0jdxt:
If have written my own script to curl the md files instead and with `bat -pP` looks pretty nice. If reading recipes in the terminal is the general use case for the curl/search it may be worth considering serving the md files in plain text.

Goblin80:
@0jdxt the idea of offering each recipe in multiple formats has already came up few times already, and personally I like it.

#158 #170 #184 

<sub>but I am not the one who calls the shots around here.<sub>

LukeSmithxyz:
As I think more of this, I'm not really sure this can be done without server side scripting. I mean it looks you're just generating a bunch of files based on words and tags and hoping someone looks for one of those. It looks like *only* single words, which I don't think anyone actually searches in the wild.

I might just want to look into a small PHP search engine that can be used on the site (and index recipe content as well), but could be extended to do this.

Goblin80:
> No one wants to see HTML

that's one more reason to consider multiple formats.

> It looks like only single words, which I don't think anyone actually searches in the wild.

I agree. it's possible to follow the same approach and consider more words in different permutations.

*but that's kinda stupid and would scale exponentially.*

---

> I'm not really sure this can be done without server side scripting

**let's take a step back and rethink the whole thing.**

1. how about delegating the search functionality to the user?

maybe something similar to this?

```bash
$ basedcooking () {lynx -dump -nonumbers -listonly https://based.cooking | fzf --preview 'lynx -dump -nonumbers -nolist {}'}
$ basedcooking
```
based-cooking-fzf-lynx-nolist-2x

I understand it's still limited by the title and tags. however its easy to build on top of that.

> and index recipe content as well

2. how about downloading all the recipes and recursively grep through them? (or something more sophisticated idk)

---

> PHP

**please please please no.**

0jdxt:
I think perhaps if we would like to keep the alias/function simple:
```bash
$ cook() { lynx -dump "https://based.cooking/$@" ;}
```
then multiple results could return the filenames e.g.
```bash
$ cook chicken curry
butter-chicken-curry
...
```

then one could do 
```bash
cook butter-chicken-curry
```
to get the recipe. And of course a single result would return recipe instead of results. Then scripters can easily pipe it into fzf or dmenu for something more sophisticated.

Goblin80:
> I might just want to look into a small PHP search engine that can be used on the site (and index recipe content as well), but could be extended to do this.

> As annoying as PHP can be it might be ok for this particular functionality since it's not complicated.

If you guys are willing to use PHP for search? then why not go full PHP??

PHP has string interpolation-ish functionality out of the box which makes generating page from templates trivial.

But then comes the question? why statically generating the site in the first place (using a sophisticated Makefile)?

Don't get me wrong, I admire all the engineering work that @sylGauthier spent into that Makefile. but there would be zero purpose for that if we went a server side rendering approach.

sylGauthier:
Here is my approach, using my homemade [menu selector](https://pedantic.software/projects/choice.html), and a generated list with the Makefile that just contains links and titles, but I will add tags as well.

output

Goblin80:
> generated list with the Makefile that just contains links and titles, but I will add tags as well.

if the list contains the links, title, tags and **content** as well, wouldn't that enable `choice` to search through the ingredients too? if so, that would be pretty good.

I ran some back of a napkin calculations, and found out that would be less than 200KB in bandwidth.

sylGauthier:
> wouldn't that enable `choice` to search through the ingredients too?

Yes that would work, but I feel like stuffing all the content into a single line is a bit overkill. It's probably easy to extract just the ingredients.

Goblin80:
>  stuffing all the content into a single line is a bit overkill

Do you mean "overkill" from a visual point of view (terminal output clutter)?

We don't have to show it, just search through it. an extra `grep`/`sed` on the recipe's content before piping everything else into `choice`.

and then after selecting, show the whole recipe.

Godje:
I'd say @sylGauthier 's idea is better than either mentioned, because the script-ability of an output like that is supreme to any other output that we have seen so far.

This also allows making multiple lists if necessary in the future to separate the content if further categories are going to be made.

@Goblin80 This essentially ends up being a big CSV file, if we're going at it. If we add any more columns to it it's pretty much a database file all in one request. Whether it's good or bad is up to you. But I'd say just keeping the title and URL in the output is a less unreasonable idea.

Goblin80:
> This essentially ends up being a big CSV file, if we're going at it. If we add any more columns to it it's pretty much a database file all in one request.

yes, precisely that.

> Whether it's good or bad is up to you.


The average recipe size (without images) is 1.2KB, without taking HTTP content encoding (gzip) nor any type of compression on server side (e.g: db.csv* as tarball).

That makes the request with other metadata (at the time of writing) about 200KB.

Loading this page to write this comment took 4.5MB.

---

I understand that might be not clever and wasteful, but on the other hand, upsides are:

* search through link, title, tag, content, ingredients, author, ...
* practically unlimited script-ability.


However, there are ways to avoid hammering the server with requests and reduce wasted traffic, one such way might be:

## `https://based.cooking/status`

This `status` endpoint is generated on build and will offer a few metadata about the current state of the site like:

* last build time
* number of recipes
* ...
* (and most importantly) a checksum of the latest state of the db.csv*

```bash
$ curl https://based.cooking/status

1616044660
119
...
ae5dd429f0549ad819958436d93c6f63 db.csv
```

and the script would honor that checksum and only fetch the whole db.csv* once (unless updated) and keep it somewhere locally.

<sub>\* it doesn't have to be csv, just anything easy to work with in grep, sed, awk, etc...</sub>

sylGauthier:
I mean ok but why not have the script just pull the whole repo at this point, it will take care of checksums and "smart update" for free.

Actually you know what, I think I might just hard fork this thing and make it into a local database of recipes. You clone the repo, you `make install`, it converts all markdown pages into man-page like files, install them in `/usr/share/cooking`, then just have a script greping stuff and displaying it correctly and call it a day.

Updates are just a periodic `git pull && make install` whenever you feel like it.

Goblin80:
> why not have the script just pull the whole repo at this point, it will take care of checksums and "smart update" for free.

I was under the impression that `git` and `make` shouldn't be a prerequisite for looking up recipes.

but on a second thought, you right I am over complicating simple stuff.
pulling the repo and running `make install` makes perfect sense.

pyzozord:
Some _thing_ has to do the search, no way around it (or is there?). The solution with pre-generating search results scales terrible and has terrible UX. I personally think that's just misguided dogma taking over the practicality resulting with the worst possible outcome. Other alternatives are:

1. **Search dynamically on the server,** with server side scripting in any number of ways (nginx + grep?) then cache the search results to save server resources. This solution adds a lot of complexity to the code, because it brings the website from static to dynamic and effectively forces admins to worry about resources and security.

2. **Search dynamically on the frontend.** Two options here, either add the fucking javascript and be part of problem instead of the the solution, or leverage browsers ctrl+f functionality by adding some static page with lots of links and descriptions (kinda crazy, but browsers can handle a lot these days, especially of extremely static, lightweight content like this).

3. If we (the server) don't want to search, and the user can't search, then the last alternative is **"make someone else search for us"**. Leverage one of the many existing web search engines, just write good robots.txt and sitemap.xml and integrate with https://duckduckgo.com/search_box or similar. This solution is really not bad, except now we depend on a third party. We better choose one that we trust and like. (Extra benefit of this solutions is that it's easy to write a nginx redirect that will redirect to search results page of a search engine.)

4. But wait, there is one more solution. DON'T DO SEARCH! :exploding_head: Essentially, the static website is a digital form of a book. Cook books are solving this problem of searching quite well I would say.

Regarding the CLI interface, it's a classic scope creep, I'd suggest you guys worry about that in a separate PR/issue. No one likes reading html source in their terminal anyway, and I doubt most people have or want to have lynx on their machines, there must be a better way to do that, if it's even necessary. Less features is worse, not better. The whole point of this project is to fight  entropy, you don't fight entropy by adding more shit.

sylGauthier:
#232 I ended up going through with my autistic solution.

Goblin80:
@LukeSmithxyz 

> So the ideal would be to run `cook chicken parmesan` and a chicken parmesan recipe is output on the terminal (provided there is only one found).


@0jdxt 

> then one could do `cook butter-chicken-masala` to get the recipe.



I have though about it for a while and came up with this:

## Leverage `comm`/`join`

The idea is find the intersection between the lists of each keyword


```bash
$ cat chicken

https://based.cooking/butter-chicken-masala.html
https://based.cooking/chicken-biscuit-potpie.html
https://based.cooking/chicken-parmesan.html
https://based.cooking/chicken-pasta-casserole.html
https://based.cooking/chicken-stock-bone-broth.html
https://based.cooking/chicken-tacos.html
https://based.cooking/chicken-tomato-spinach-curry.html
https://based.cooking/chicken-wings.html
https://based.cooking/japanese-noodle-soup.html
https://based.cooking/pan-seared-chicken.html
https://based.cooking/roasted-chicken-breast.html
https://based.cooking/spatchcock-chicken.html
https://based.cooking/stir-fried-chicken-with-an-orange-sauce.html

$ cat parmesan

https://based.cooking/chicken-parmesan.html
https://based.cooking/parmesan-potatoes.html

$ join chicken parmesan

https://based.cooking/chicken-parmesan.html

$ cat butter | join - chicken | join - masala

https://based.cooking/butter-chicken-masala.html
```

building on that idea, this snippet uses `curl` instead of `cat`:


```bash

$ basedcooking() {
URL="https:\/\/based.cooking\/"
curl="curl -sL $URL"
eval `echo "$@" | sed -re "s/ (\w+)/ \| join - <(echo \"\$\($curl\1\)\")/g" -e "s/^/$curl/"` | fzf --preview 'lynx -dump -nonumbers -nolist {}'
}

$ basedcooking chicken parmesan
```

based-cooking-search-join-opt


<sub><sub><sub>eval?! cringe</sub></sub></sub>

## Recipe's content

Instead of considering the title and the tags only, the search index can be extended to include individual words in the recipe, which **enables searching contents** of a recipe.

Each word in the recipe's content will be treated as a keyword. same as the words in the title or the words in the tags.

The search index would be about 3k words/files max.

---

@pyzozord

> The solution with pre-generating search results scales terrible and has terrible UX.

In my defense, if you read the **curl/Search function in the future** section in the README.md, you will find this PR is trying to solve exactly that.

> Search dynamically on the server, with server side scripting in any number of ways (nginx + grep?) then cache the search results to save server resources. This solution adds a lot of complexity to the code, because it brings the website from static to dynamic and effectively forces admins to worry about resources and security.

100% agree.

> just write good robots.txt and sitemap.xml

yea, good point. the site is missing robots.txt and security.txt


https://based.cooking/robots.txt
https://based.cooking/.well-know/security.txt


see: https://securitytxt.org/

> No one likes reading html source in their terminal anyway, and I doubt most people have or want to have lynx on their machines.

That's why I opened issue #219


> you don't fight entropy by adding more shit.

You are absolutely right.

---


## In Conclusion

@Godje @LukeSmithxyz

Hands down, @sylGauthier's "autistic" solution is light years better than all the gimmicks that I am trying do in this PR.

If you went ahead and merged his PR #232, please close mine. I don't think I can do any better.

pyzozord:
@Goblin80 kudos for the sober attitude. Things we don't do have just as big, if not bigger, impact than the things we do. It's way more difficult to decide what not to do.

> In my defense, if you read the curl/Search function in the future section in the README.md, you will find this PR is trying to solve exactly that.

Oh I actually didn't notice that. :man_facepalming: @LukeSmithxyz I think that's a really bad design decision. Cooking recipes in command line are a gimmick. Instead you can have a prominent section in readme.md with "satellite" repos with useful scripts and forks that don't fit to the primary functionality.

Also think again about the search. Curated lists, index and tags are way more based. Search is for soyboys.

sylGauthier:
> Cooking recipes in command line are a gimmick

IMHO running a huge, bloated, millions-lines-of-code javascript sandbox aka modern web browser just to render 10 lines of instructions on how to mix ingredients is the real gimmick.

I'd say ideally, there would be one repo *only* for markdown recipes, a repo with the tools to generate the based.cooking website then as you said, other satellites repos with tools to convert the recipes into whatever floats people's boats. That would be the basedest solution.

Also this would allow @LukeSmithxyz to give (some) people collaborator status to the recipe repo so that a small team of moderators can review/merge recipes, while keeping control over the website's internal tools, CSS, templates, etc.

DominikRusso:
I'm still not quite sure why we need a curl interface when the recipes are in a git repo. Isn't it much easier and cheaper to just clone it and use whatever command line tools you'd like to search locally?

sylGauthier:
> Isn't it much easier and cheaper to just clone it and use whatever command line tools you'd like to search locally?

Yes. #249 #232

SSouper:
This project is in a dire need of database. (maybe keep it portable and use sqlite + nginx caching?)
That way you can actually implement a decent search function, pagination(the index page will just keep getting slower and slower to load) and possibly including recipes in other recipes.

uKruel:
A simple Front End solution would be using a HTML <form> with a <datalist> and some <option> html tags with values that have cooking tags and recipe names. It just uses HTML with a HTTP GET request. 

[Here is an example](https://www.w3schools.com/TAGs/tryit.asp?filename=tryhtml5_datalist)

DominikRusso:
> the index page will just keep getting slower and slower to load

We're at least 10 000 recipes short of the index page being too large to load quickly.

RidhwaanDev:
Search is bloat by the way. You don't need it. Just use command-f. That is the only way.

sylGauthier:
For those interested in man page recipes, I made a script to convert/install/read/search markdown recipes from any source. You can just clone this repo, run:
```
cook -i src/*.md
```
to install all recipes and that's it. Available [here](https://github.com/sylGauthier/cook).

You can do
```
cook -s "word"
```
to grep for `word` in all recipes, something that you won't get with ctrl-f on the web-based solution without filthy PHP or javascript.

Godje:
I agree with @sylGauthier that this whole thing has to be fragmented to keep everyone's sanity. https://github.com/LukeSmithxyz/based.cooking/issues/249

We can have people maintaining a satellite website repo that will generate the webpages and stuff, and we can have the people maintaining the Markdown list of recipes that has a strict ruleset.

This way people who want to have the whole recipe book in a terminal should either clone the MD repo and have their own based tool, or come together on a separate command line tool that will fetch the MD list and do all their stuff automatically, considering some people just want to see a gosh darn recipe, and not participate in this autistic pursuit of a perfect technological masterpiece.

PR #167, opened 2021-03-15
fix RSS titles, put whole recipe in description

sylGauthier:


LukeSmithxyz:
Well, they still need to be in HTML/run through discount to be readable by an RSS reader.

PR #172, opened 2021-03-15
Header now links to homepage

remnny:
because i find myself clicking on it and being disappointed that it isn't a link


LukeSmithxyz:
Found the r*dditor.

I don't really like this at all. I think a link to the index at the top is a good idea (maybe in the form of a small icon), but it makes no sense to me for someone to click on the name of the page and that takes you to the homepage. That's a design feature as annoying as R*ddit or Tw*tter.

LukeSmithxyz:
My caprices have changed and I've decided this is okay. I think it's better than having some extra line stolen by links.

LukeSmithxyz:
Crap, actually, I'm going to revert this because it doesn't work with the markdown converter.

More reason we need a new static site generator.

PR #177, opened 2021-03-15
spelling conform to new title

siggines:
England-English spelling conformity in suit with title change. ( from 'pork based chili-con-carne' to 'chilli-con-carne', presumably chosen to distinguish it from the other chilli recipe without need of a long title. ) It just bugs me that the spelling is 'Chilli' but the use of it on the page is 'Chili'.


LukeSmithxyz:
> tomatoe
do bongs really?

PR #187, opened 2021-03-16
Fixed Image sizing for directional images

hallzy:
The `li img` selector looks like it was meant to target the bitcoin and monero icons on the homepage, but it is also targeting images that people have put at certain steps of recipe directions.

I have Updated the css selector to be more specific so that it won't target directional images on recipe pages anymore.

Example page: https://based.cooking/stuffed-round-squash (note the really small images at the end of directions 1 and 2).

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -25,7 +25,8 @@ code {
 	color: lime ;
 }
 
-li img {
+img[alt="BTC logo"],
+img[alt="XMR Logo"] {
 	max-width: 1em ;
 	max-height: 1em ;
 	display: inline ;


LukeSmithxyz:
Well, honestly, I don't really feel that the submitter did the best thing by including the images in the list by themselves...

But I suppose there is another problem in `discount` as a markdown converter: it doesn't follow ordered lists that are separated. E.g. we *should* be doing this:

```
1. Do this.

An image

2. Now do this.
```

But `discount` counts replaces the html list starting with `2.` with `1.` apparently (lowdown did not do this). I suppose there has to be a way around this though.

PR #198, opened 2021-03-17
Correct grammar and name

throwaway1037:
"Parmesan" is a generic name for the cheese Parmigiano-Regiano. I propose instating the rule that this traditional, authentic cheese must always be preferred over the cheaper, mass-produced "parmesan".

Please see: https://en.wikipedia.org/wiki/Parmigiano-Reggiano


LukeSmithxyz:
Lol. Well I guess that would be "based." I think that most people automatically all that into the word "Parmesan" anyway (at least the people who would care enough to not get the cheap stuff) but whatever.

PR #205, opened 2021-03-17
Remove bottom margin

m-chrzan:
Was there any reason for a big margin at the bottom? Pages look cleaner this way in my opinion.

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -2,7 +2,6 @@ body {
 	max-width: 800px ;
 	margin: auto ;
 	padding: 0 16px ;
-	margin-bottom: 500px ;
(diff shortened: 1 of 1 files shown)


plainoldcheese:
I also agree

siggines:
Ah, but you see when you scroll down to the bottom, with that extra space, the last text of the screen... is centre screen. This is good aesthetics.

Farioko:
Wasting half of the screen real estate after jumping to the bottom of the page is NOT good aesthetics. From a usability perspective this is horrible, it violates multiple design principles. The scroll bar indicates to the user that there is more content, whilst there isn't.

PR #211, opened 2021-03-17
Add index link to top of article

bigtimes:
In reference to #172, where a suggestion was made to have a link at the top to return to the index of recipes. 

Here's how it looks: 
retvrn

### DIFF
--- a/templates/article_header.html
+++ b/templates/article_header.html
@@ -1,0 +1,1 @@
+<p><a href="/">⟵ Recipe index</a><p>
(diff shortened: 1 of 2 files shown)


metruzanca:
Personally, I liked the changes you made in #172, as that's a UX feature many come to expect nowadays. And its a pretty convenient way to have a link back to home without 1. adding a button/link or 2. having to scroll to the bottom to find the "homepage" button (which nobody expects to be there).

_Luke is probably gonna tear me a new one here after I said that..._

---

That said, I also like this alternative.

Aesthetically, maybe remove the underline in the arrow?
```html
 <p>⟵ <a href="/">Recipe Index</a></p>
 ```

PR #221, opened 2021-03-18
Make the look and feel a bit warmer.

el3ctr0lyte:
I changed the white color for the text to navajowhite which is a bit warmer. I changed the link color to gold, to fit the theme of the website a bit better. And lastly I added a bottom border to the `<h1>` tags, as a way of adding a horizontal rule that looks pretty nice.
Screenshot_2021-03-18_16-39-41
This is what it looks like.

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -1,6 +1,6 @@
 body {
 	background: #151515 ;
-	color: white ;
+	color: navajowhite ;
 	max-width: 800px ;
 	margin: auto ;
 	padding: 0 16px ;
@@ -9,15 +9,17 @@ body {
 }
 
 a {
-	color: lightblue ;
+	color: gold ;
 }
 
 a:visited {
-	color: gray ;
+	color: darkgoldenrod ;
 }
 
 h1 {
 	text-align: center ;
+	border-style:solid;
+	border-width:0px 0px 2px 0px;
 }
 
 h2 {


el3ctr0lyte:
Warmer colors are psychologically more `comfy`, apperantly.

el3ctr0lyte:
Just ignore everything after the comment about `warmer color are more comfy`. I was retarded and wanted to add a recipe but i didn't know that those commits would show up in this pull request.

metruzanca:
@el3ctr0lyte, I like these styling changes. Look easier on the eyes. Yellow light and all vs blue light.

Though you may want to undo those commits by checking out `e68a029` (Update style.css) then force pushing on your branch. That should get rid of those 6 commits and will also make it so when merging those won't show up in the git history. (though I presume luke would squash them anyways).

If you wanna add a recipe, just do `git checkout -b my-recipe-name-goes-here` then make a second PR.

el3ctr0lyte:
Thanks @metruzanca! I only know very basic git commands and for these recipes i used github's interface to make styling changes and all. I would like very much to remove those commits. Could you guide me in the right direction? Is there a way to do this from the web interface?

el3ctr0lyte:
Nevermind, googled it and figured it out!. Thanks!

metruzanca:
@el3ctr0lyte I'm Glad I could help in some way, awesome that you figured it out!

Now we just gotta get this branch merged by a Luuuke haha

PR #232, opened 2021-03-19
add rule to generate/install local manpages from markdown recipes

sylGauthier:
- Add a rule in the makefile to generate `pages/%.1`, groff files using the `man` macro package (because I was lazy and it just works).
- Install them using `make install` wherever `COOKPATH` points to in the `config` file.
- The converter is just a *very* basic sed script embedded in the Makefile so no deps required.
- All relative links in markdown files are expected to be links to other recipes (which makes sense), but the corollary is that all links to other recipes *must* be relative so that the converter uses the proper formatting. So this PR comes with a few fixes in some recipes, and it makes sense to have a relative linking system anyways since people need to be able to host their own instances of this site.
- Add `cook`, a small, proof-of-concept script to navigate/search through the installed recipes. It requires to set `$COOKPATH` eg. in your `bashrc`, similarly to the `$MANPATH` variable.
- The idea is that `cook` should be a standalone, separate tool, and recipes can be installed from different sources.
- `cook` works recursively from the `$COOKPATH` path, which allows to install recipes in subdirectories so that different sources can have similar recipe names.

Here is a demo of what it would look like to have `cook` already installed on your system, if you wanted to pull and install based.cooking recipes:

output

I would totally understand if this is considered too far off the initial goal and gets rejected, but in this case I will definitely maintain my fork up-to-date because this is the way I want to browse my recipes.


sylGauthier:
Closing PR since my tool is now completely standalone and can install single markdown recipes into manpage itself.

Get the tool [here](https://github.com/sylGauthier/cook).

PR #241, opened 2021-03-19
Suggestion to increase line-height

taco-c:
An increased line-height to make it easier for mobile users to click on the links. Also as a bonus makes blocks of text easier to read, both on desktop *and* mobile. The value of 1.5 is good in my opinion, but 1.75 or similar could also work.

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -6,7 +6,6 @@ body {
 	padding: 0 16px ;
 	margin-bottom: 500px ;
 	font-family: sans-serif ;
-	line-height: 1.5 ;
 }
 
 a {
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
1.5 is going to be hugely ugly and of course will make the website 150% of the size.

If we do include something like this, it should be:

1) *only* on mobile
2) *only* for acute situations where it might be necessary (i.e. between list-item links)

Frankly even then, I'm skeptical.

taco-c:
Changed to:

1. only for devices with imprecise pointer control, i.e. touch devices
2. only on `li` elements

See if you like it.

Issue #3, opened 2021-03-10
Need a Logo and a faveicon


LukeSmithxyz:
I want a little logo, just for the faveicon for now.

If anyone can think up anything creative, tell me. Otherwise I'll make something up. Should be something based.

If it's a simple design, it should be a native svg.

lisbonjoker:
![image](https://user-images.githubusercontent.com/80425961/110695290-97d6e600-81e1-11eb-822f-ea938f4c1899.png)
Simple?

LukeSmithxyz:
Well I'd really just like a square image only, preferably one a little colorful or notable for the faveicon.

I'm not actually going to put this on the page itself now (nor do I really plan wasting the space).

lisbonjoker:
I'm going to send the SVG just doing a few drafts to see what you think of them
![image](https://user-images.githubusercontent.com/80425961/110696781-6828dd80-81e3-11eb-838e-5ed1a58b0363.png)
Colorful

LukeSmithxyz:
Well, when I mean "colorful," I don't mean monochromatic.

lisbonjoker:
Oh mistake, sorry

lisbonjoker:
Any color suggestion to put on it?

lisbonjoker:
Earthly color?
![default-monochrome](https://user-images.githubusercontent.com/80425961/110698304-63652900-81e5-11eb-804f-b6b9744b6088.png)

LukeSmithxyz:
Lol. Monochomatic = one color. It's still one color.

Example of something non-monochromatic: A gray pan, a colored handle for the spatula, multiple ingredients jumping out of the pan all of multiple colors.

lisbonjoker:
Oh I'm dumb sorry, it's the lack of sleep lately

beune:
Maybe generate one using [https://favicon.io/](https://favicon.io/)?

twiclo:
I like the skillet in the title for the faveicon: 🍳

fbosch:
> I like the skillet in the title for the faveicon: 🍳

you can use emoji as favicons directly -

https://css-tricks.com/emojis-as-favicons/

lisbonjoker:
![default-nonmonochrome](https://user-images.githubusercontent.com/80425961/110703073-39166a00-81eb-11eb-8a69-8d1642e3b87c.png)
Fourth time's the charm

twiclo:
[...]

PR #256, opened 2021-03-20
Add a lightweight inline favicon

kigiri:
### Issue
Browser will query for a `favicon.ico` if none were specified.

image

### Solution
Using emoji and SVG we can create a **very light** favicon that is still descriptive of the website and avoid having one more request done to the server for nothing.

### DIFF
--- a/templates/header.html
+++ b/templates/header.html
@@ -3,6 +3,7 @@
     <head>
         <meta charset="UTF-8">
         <meta name="viewport" content="width=device-width, initial-scale=1">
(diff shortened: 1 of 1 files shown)


teunissenstefan:
I second this. If you don't supply a favicon, the browser will still look for one. This takes around 120ms on my system. By using an SVG favicon, this can be greatly reduced.

kigiri:
I just saw https://github.com/LukeSmithxyz/based.cooking/issues/3 about needing a favicon and svg logo, this can still help in the meanwhile, and when you get a real logo you can replace it in the code instead of the emoji.

PR #257, opened 2021-03-20
Fix Brigadeiro recipe capitalization and added a picture

teunissenstefan:
I added a picture, changed the capitalization of "Brazilian", and added the "brazilian" tag.


Knarkzel:
Good.

teunissenstefan:
Before (including an empty last page because of the margin on the body):
image

After:
image

PR #263, opened 2021-03-20
Create Minimalist_Egg.md

marttoygoy4:
A very minimalist egg.


LukeSmithxyz:
There's already an egg page with multiple recipes.

PR #270, opened 2021-03-20
A quick modification

actuallyexeon:
1. Added one more tag
2. Added optional step
3. Fixed some grammer


actuallyexeon:
It seems like lots of deploys have been broken after this... maybe it was the tag I added?

PR #272, opened 2021-03-20
Created Lasagna.md

fres7h:
thought based.cooking needed a lasagna recipe. <3


LukeSmithxyz:
Could you include the serving/prep info in a list with emojis like the example? I know not all pages have this yet, but I want to keep it standard.

fres7h:
fixed! since you're here, i was meaning to ask if you took BCH for donations, i'm aware it's a shitcoin but the fees are low and i'm basically broke.

PR #277, opened 2021-03-20
Add workflow to publish to github pages

Joshix-1:
This workflow builds the website to a GitHub pages branch.

For it to work you just need to add keys to the repo (see: https://github.com/peaceiris/actions-gh-pages#%EF%B8%8F-create-ssh-deploy-key )

Then the page can be hosted with GitHub pages, and it doesn't cost you anything.
In the settings of the repo you can then specify a domain to publish the page to.

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -1,0 +1,34 @@
+name: Deploy to GitHub Pages
+on:
+    push:
+        branches:
+            - master
+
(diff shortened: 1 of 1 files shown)


Knarkzel:
Don't think this is necessary.

Joshix-1:
It would reduce hosting costs to 0.

PR #279, opened 2021-03-20
Breakfast hash, super bacon

nephitejnf:
A pretty good regular breakfast regimen.


Knarkzel:
Nice, please add prep etc:

```
- ⏲️ Prep time: 10 min
- 🍳Cook time: 30 min
- 🍽️ Servings: 4
```

PR #281, opened 2021-03-20
Add images size checking in PR

danielleontiev:
Workflow to check images size in PR


marttoygoy4 (commented):
I support the idea :)

.github/workflows/scripts/check-size.sh
```diff
@@ -0,0 +1,25 @@
+#!/bin/bash
+set -eu
+
+SIZE_LIMIT=150000
```
zyansheep, line 4: Set limit to 15360? (16k resolution) 150000 is too high I think

PR #289, opened 2021-03-20
Hide footer and remove margin in print view

teunissenstefan:
For a better looking print page


zyansheep:
Can you provide images?

teunissenstefan:
Before (including an empty last page because of the margin on the body):
image

After:
image

PR #302, opened 2021-03-20
Update CSS

frogtd:
This adds padding when the pointer is course, like on a phone to, so people can more easily click on a recipe.

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -65,7 +65,12 @@ img[alt="XMR Logo"] {
 		color: forestgreen ;
 	}
 }
-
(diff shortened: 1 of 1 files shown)


data/style.css
```diff
@@ -65,7 +65,12 @@ img[alt="XMR Logo"] {
 		color: forestgreen ;
 	}
 }
-
+@media (pointer: coarse) {
+	li > a {
+	    display: block;
+		height: 1.5rem
```
hallzy, line 71:
```suggestion
	height: 1.5rem;
```

PR #303, opened 2021-03-20
Apple Strudel recipe

lorenzoiuri:


zyansheep (approved):
Nice photos! the presentation is very well done.
Other than that, does Luke want images to be formated with a leading 0? (e.g. apple-strudel-01.webp)

Knarkzel:
```
If you would like to add additional directional images, they should be numbered with two digits like: pix/chicken-parmesan-01.webp, etc.
```
The images are numbered correctly, leading zeroes are unnecessary.

PR #304, opened 2021-03-20
Added soy latte recipe.

jirian:
Added soy latte recipe.


DickFaggotson:
The thing has soy in its very name and yet you submit it, lol.

jirian:
@DickFaggotson Pun intended!

Morelcia:
so true

sylGauthier:
DiLatte.

PR #307, opened 2021-03-20
add eggroll in a bowl recipe

the-birchmen:

### DIFF
--- a/src/eggroll-in-a-bowl.md
+++ b/src/eggroll-in-a-bowl.md
@@ -1,0 +1,34 @@
+# Egg Roll in a Bowl
+
+Egg rolls, but, you know, in a bowl.
+
+- ⏲️ Prep time: 10 min
+- 🍳Cook time: 15 min
+- 🍽️ Serving 22
+
+## Ingredients
+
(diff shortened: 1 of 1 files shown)


src/eggroll-in-a-bowl.md
```diff
@@ -0,0 +1,34 @@
+# Egg Roll in a Bowl
+
+Egg rolls, but, you know, in a bowl.
+
+- ⏲️ Prep time: 10 min
+- 🍳Cook time: 15 min
```
hallzy, line 6:
```suggestion
- 🍳 Cook time: 15 min
```

src/eggroll-in-a-bowl.md
```diff
@@ -0,0 +1,34 @@
...
+- 🍽️ Serving 22
```
hallzy, line 7: Could you elaborate on what `22` means?

the-birchmen: No it is 2 servings. I was having issues with markdown rendering in vim. I think it has something to do with emojis. It looked right when I pushed it. Apologies.

PR #318, opened 2021-03-21
Add Baked Mostaccioli Recipe

zyansheep:
Here is my decade-old family pasta dish
Hope I didn't miss anything!


sylGauthier:
Maybe rename in `mostaccioli-baked.md` so that it's ranked in "M" on the front page (ordering is by file name).

Add tag `italian`, and I would say, remove commercial links. People can figure out how to search themselves if they need to.

zyansheep:
I added the italian tag, but I think i'll keep it `baked-mostaccioli` because there are a lot of other recipes that start with `baked-`

PR #321, opened 2021-03-21
style.css: increase spacing

wesleydeal:
Increase text readability by adding a bit to the line height and a slight gap between columns on the front page. This makes it much easier to read through the whole list of recipes without the words running together.

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -6,6 +6,7 @@ body {
 	padding: 0 16px ;
 	margin-bottom: 500px ;
 	font-family: sans-serif ;
+	line-height: 1.4 ;
 }
 
 a {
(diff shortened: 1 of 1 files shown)


wesleydeal:
Screenshots for comparison:
(proposed)
image

(current)
image

taco-c:
I have a similar pull request at #241.

wesleydeal:
Ah, I didn't see that. I think the readability concern is perhaps just as prevalent on desktop as the potential for misclicking on mobile... perhaps line-height: 1.3 or so on desktop would be a decent compromise with Luke's page size concern? With perhaps a little extra spacing in the menus on mobile devices.

taco-c:
I don't know what browser you use, but most browsers use a default line-height of 1.2-ish, so I'm not sure how much difference 1.3 would really make. And in my opinion 1.3 is still too small for mobile misclicking prevention; 1.5 is the lowest I would go in that regard, else it's not worth the trouble.

PR #329, opened 2021-03-21
[Changed at styles.css] Huge, I mean MASSIVE images ruin the exprience

marttoygoy4:
image

----------------------------
After the my changing it should look like:
image

I think adding a 'checkbox' that let you change the pictures from max-width-250px to 600px 
could be nice too, but I tried it on my smart phone and the images look actually sharper and I was able to follow the recipie better! 
Please consider my change

here is a good page you can try the change 
```https://based.cooking/arroz-chaufa```

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -29,7 +29,8 @@ footer {
 }
 
 img {
-	max-width: 600px ;
+	padding-left: 120px;
+	max-width: 250px ;
 	width: 100% ;
 	margin: auto ;
 	display: block ;

### CODE REVIEW

**Comment by Knarkzel** [comment]
Why do you use absolute values?

```css
padding-left: 120px;
max-width: 250px ;
```

data/style.css
```diff
@@ -29,7 +29,8 @@ footer {
 }
 
 img {
-	max-width: 600px ;
+	padding-left: 120px;
+	max-width: 250px ;
```
zyansheep, line 33:
```suggestion
	max-width: 50%;
```
120px shows up tiny on my monitor. I think 50% of page is probably what you want, but I'm not sure if other people think the same.

Justin-CB: I'm not sure about the exact value, but I agree that having the width value be a percentage instead of an absolute pixel value does make a lot of sense. If we wanted to have the maximum value be an absolute pixel width, we should enforce that as a condition for submitting images instead of in the CSS.

marttoygoy4: MMmmmh if you wanted why not using the resizer tool chrome provide we ahd it in windows-xp

Hate9: The only issue with using a percentage is that mobile will have issues. 
Personally, I think we should go with a percentage-based max width (or maybe even just width) for images, and then a media rule to expand it in portrait orientations such as:
`
img {
    max-width: 60%;
}
@media screen and (orientation:portrait) {
    img {
        max-width: 85%;
    }
}`
with screen specified so we don't mess with printers.

PR #339, opened 2021-03-21
add meta description tag

Odyssey346:
This will get the site a better SEO score in Lighthouse. I believe it's used by search engines to show a description for the site? Good nonetheless. (if it should be changed lemme know)


Knarkzel:
Looks good to me. See [meta tags](https://www.metatags.org/all-meta-tags-overview/the-important-meta-tags/meta-name-description/) for why.

PR #341, opened 2021-03-21
[Change in index.md] Please considering chaging the approch

marttoygoy4:
Seemly for me a way nicer approch


Knarkzel:
This change is moronic.

marttoygoy4:
> This change is moronic.

Eh I generally think they way he says it, is a bit aggresive

PR #343, opened 2021-03-21
Cleaner CSS scheme

PedroEdiaz:
It's just less bloated to make a query only for the dark colour scheme.

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -1,6 +1,4 @@
 body {
-	background: #151515 ;
-	color: white ;
 	max-width: 800px ;
 	margin: auto ;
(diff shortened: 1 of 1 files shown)


Knarkzel:
I'm not sure if I agree with this change, although it's nice to have less code
in the repository. Changing it such that the user MUST have `prefer-color-scheme: dark`
in order to view the (already agreed upon default that everything should be dark mode) makes
it more annoying for those of us who may not have set this property to dark, such as on mobile.

PedroEdiaz:
I didn't notice that, you got a point there.

PR #372, opened 2021-03-21
Specialty Grilled Cheese

bradenorchard:


zyansheep (changes requested):
Wow, that's a fancy grilled cheese recipe! I think it would be good to generalize it for optional complexity or give it a specific name. imho the grilled-cheese.md recipe should be simple.

Knarkzel:
Missing following:

```
- ⏲️ Prep time: 10 min
- 🍳Cook time: 30 min
- 🍽️ Servings: 4
```

bradenorchard:
Eh, never mind then.

PR #374, opened 2021-03-21
Add recipe for kvass

Holsterbau:
[Kvass](https://en.wikipedia.org/wiki/Kvass) is a Slavic fermented beverage which is both simple to make and rich in flavour.

### DIFF
--- a/src/kvass.md
+++ b/src/kvass.md
@@ -3,7 +3,7 @@
 [Kvass](https://en.wikipedia.org/wiki/Kvass) is a Slavic fermented beverage which is both simple to make and rich in flavour.
 
 - ⏲️ Prep time: ~30 min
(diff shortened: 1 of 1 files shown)


src/kvass.md
```diff
@@ -0,0 +1,38 @@
+# Kvass
+
+[Kvass](https://en.wikipedia.org/wiki/Kvass) is a Slavic fermented beverage which is both simple to make and rich in flavour.
+
+- ⏲️ Prep time: ~30 min
+- 🍳 Cook time: ~1 week
```
zyansheep, line 6:
```suggestion
- ⌛ Fermentation time: ~1 week
```

zyansheep (changes requested):

PR #378, opened 2021-03-21
Add image for pancake recipe

abbasmaheryar:
I'm not that much of what you call a cook. The pancakes turned out alright; at least that's what I thought. I had to make some conversions to freedom units, and I think I had to add some extra flour at the end since the batter was very thin.


Knarkzel:
The pancake image is very large at 761kb, you should resize it to atleast below 200kb.

zyansheep (approved):

abbasmaheryar:
i reduced the file size to 177 kb. i'm not sure why the check is failing

Holsterbau:
@abbasmaheryar according to the workflow script, <=150 kB is indeed the limit.

PR #380, opened 2021-03-21
Quesadilla recipe improvements + add picture

zyansheep:
Made the recipe more streamlined, modular, and easier to understand. Also added a picture.


src/quesadilla.md
```diff
@@ -1,30 +1,35 @@
 # Quesadilla
 
-A quick and easy meal, able to be taken anywhere, but be warned of its high empty carb counts.
+![brisket quesadilla](pix/quesadilla.webp "Brisket and Bean Quesadilla with using Montery Jack Cheese")
```
Knarkzel, line 3: Correct me if I'm wrong but is this valid image syntax? description

zyansheep: Yes it is, the part in quotes represents what you see when you mouse over it https://raw.githubusercontent.com/LukeSmithxyz/based.cooking/master/src/baked-mostaccioli.md

Knarkzel: Hmm I see, wonder if this standard should be enforced once we start cleaning up...

zyansheep: It's too obscure I think. I would leave it optional. The only time people use that feature is for xkcd comics...

Issue #354, opened 2021-03-21
Tag system bug


throwaway1037:
Some recipes are repeated twice when you visit a particular tag.

Eg. [the cake tag](https://based.cooking/@cake).

hallzy:
Interesting. I wonder if at least in the case of the `cake` tag that it is because it uses the `pancake` tag also as a `cake` tag.

Something that the duplicates have in common for that tag are that they are tagged as both `pancake` and as `cake`.

throwaway1037:
I hate to say it, but there are still duplicates.

hallzy:
@throwaway1037 It doesn't look like the PR is actually live yet. I check a recipe change before and after this, and neither recipe is changed on the site either, so I am going to assume that that is the reason for now.

throwaway1037:
It's fixed,

hallzy:
@throwaway1037 ah I was trying to find this issue so I could reference it

I noticed that the CI was broken #413 and that has been fixed now, so that's why nothing was updating.

PR #385, opened 2021-03-21
Fixes duplicates on tag pages (#354)

hallzy:
See issue #354. A tag page like `cake` was also using recipes tagged as `pancake` because it contained the word `cake`. Since some pancake recipes were also tagged as cake, they appeared twice.

Solution was to specify that grep look for a line that is the name of a tag, instead of seeing if the line contains the name of a tag.


hallzy:
I updated this PR to use `awk` instead of `grep | sed` now. After removing the `-F`.

hallzy:
@sylGauthier Thought I'd include you in on this since you work on blogit.

Issue #365, opened 2021-03-21
compound tags?


smantha69:
It seems my [recent PR](https://github.com/LukeSmithxyz/based.cooking/pull/315) has introduced a bug. It doesn't look like hyphenated tags are supported. `slow-cooked` has been broken into [`slow`](https://based.cooking/@slow) & [`cooked`](https://based.cooking/@cooked).

Here's a few ideas for handling this: 
* omit it entirely: `slowcooked`
* just use `slow`
* add support for this (least attractive)
* use some other supported separator

PR #386, opened 2021-03-21
Fix tag separator to allow hyphen separated tags like 'slow-cooked'

hallzy:
Consider this [chicken-tacos](https://based.cooking/chicken-tacos) recipe.

it has the `slow-cooked` tag on it. But if you click the tag on this page you get taken to a page that doesn't exist because the tagging separates on punctuation, which includes hyphens. So you actually end up with this recipe in a tag called `slow` and a tag called `cooked`.

I changed the `tr` command to use the `-c` option for the complement of SET1, and SET1 is now alpha characters and hyphens. So now tags will be split on characters that aren't alpha or hyphens, which should work for any case we have.

I decided not to do capital letters, because tags shouldn't be capitalized anyways.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -59,7 +59,7 @@ config:
 
 tags/%: $(BLOG_SRC)/%.md
 	mkdir -p tags
-	grep -ih '^; *tags:' "$<" | cut -d: -f2- | tr '[:punct:]' ' ' | sed 's/  */\n/g' | sed '/^$$/d' | sort -u > $@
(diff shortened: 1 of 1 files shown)


hallzy:
REF #365

The slow-cooked recipes aren't the only ones. There is also a tex-mex, baking-pan, and smoked-rolled-roast

zyansheep (approved):
the regex looks ok to me, I haven't tested it though

hallzy:
@sylGauthier Thought I'd include you in on this since you work on blogit.

PR #392, opened 2021-03-21
Glutenfree Pfannkuchen: Correct tags format

benarmstead:


src/glutenfree-pfannkuchen.md
```diff
@@ -32,4 +32,4 @@ A very easy recipe for glutenfree Pfannkuchen (german pancakes).
 
 - Lxgr - [github](https://github.com/lxgr-linux)
 
-;tags: basic, sweet, breakfast, german, pancake
+;tags: basic sweet breakfast german pancake
```
Knarkzel, line 35: This is correct format, but you should only have at most 2-3 tags.

PR #397, opened 2021-03-21
Update pull_request_template.md

hallzy:
I've noticed still some PRs are coming in that don't comply with the PR template.

I've seen other projects do similar things that I have attempted here which is to require that someone checks the boxes to make sure they actually read the thing, and then it shows up for everyone else as well so we can see that they have read it.

Not sure if we want to take this approach, but thought i would at least make it an option.

Below is how it would look:

- [ ] I am adding or changing a recipe and I understand and agree to the below
  statements:
    - [ ] My pull request may be closed without reason or notice if:
        - My PR does not conform to all of these rules,
        - I have left one or more of these rules unchecked, or
        - This checklist is missing from my PR.
    - [ ] My recipe is a `.md` file in the `src/` directory (refer to existing
      recipe files or the [example](example.md)).
    - [ ] The filename for my recipe is:
        - The name of the dish,
        - Written in all lowercase, and
        - Each word is separated by a hyphen (`-`)
            - Good: `chicken-tacos.md`
            - bad: `Chicken-Tacos.md` (has capital letters)
            - bad: `chicken tacos.md` (uses spaces)
            - bad: `chicken_tacos.md` (uses underscores)
            - bad: `my-first-recipe.md` (does not describe what the recipe is for)
    - [ ] My recipe is based, i.e. good traditional and substantial food.
      Nothing ironic, meme-tier hyper-sugary, meat-substitute, etc.
        - Meat substitutes are fine if they are not being used for the purpose
          of replacing meat. ie if your recipe is a traditional recipe that uses
          tofu then this is fine.
    - [ ] My recipe contains 2 to 5 relevant tags at the bottom of the recipe
      file. Try to use already existing tags.
    - [ ] My recipe doesn't include salt and pepper and other ubiquitous things
      in the ingredients list.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,12 +1,26 @@
-<!--
-- Recipes should be `.md` files in the `src/` directory.  Look at already
-  existing `.md` files for examples or see [example](example.md).
-- File names should be the name of the dish with words separated by hyphens
-  (`-`). Not underscores, and definitely not spaces.
-- Recipe must be based, i.e. good traditional and substantial food. Nothing
-  ironic, meme-tier hyper-sugary, meat-substitute, etc.
-- Be sure to add a small number of relevant  tags to your recipe (and remove
-  the dummy tags). Try to use already existing tags.
-- Don't include salt and pepper and other ubiquitous things in the ingredients
-  list.
--->
+- [ ] I am adding or changing a recipe and I understand and agree to the below
+  statements:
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
This is just way too long. The comments I have already are frankly too long. It needs to be concise and non-overwhelming. The scripts will already check for most of this stuff.

hallzy:
Pretty sure I added this PR before I updated the script to actually check for some of these things, so I'll just close this.

PR #400, opened 2021-03-22
Updated PR workflow to check more things

hallzy:
tbh, I've never written a github workflow before, but seems like this should work. I don't really know how you are supposed to test github workflows without having them on github first.

I ran some tests running the script manually with specific files and it is working like that.

Basically added checks to make sure that certain sections exist, and we have a decent number of tags in each submission.

Also excluded files in the .github folder from being checked.


.github/workflows/scripts/check-files.sh
```diff
@@ -0,0 +1,125 @@
...
+
+git diff --name-only "$(git merge-base origin/master HEAD)" | while IFS= read -r file; do
+	case "$file" in
+		*.webp)
+			echo "Checking size of $file"
+			check_size "$file"
+			check_webp_name "$file"
+			;;
+		.github/*.md)
+			exit 0;
```
danielleontiev, line 118: Looks like it will exit as soon as it meets .md file in .github subdirectory leaving other files uncheked

.github/workflows/scripts/check-files.sh
```diff
@@ -0,0 +1,125 @@
...
+
+check_size() {
+	size="$(stat --printf="%s" "$1")"
+	if [ "$size" -gt "$SIZE_LIMIT" ]; then
+		echo "File $1 is bigger than specified $SIZE_LIMIT limit"
+		exit 1
+	fi
+}
+
+check_webp_name() {
```
danielleontiev, line 14: The only difference between this func and func below is the regular expression used. I think both functions could be unified taking regex as a parameter

.github/workflows/scripts/check-files.sh
hallzy: This is the intent. The markdown file in the .github folder shouldn't be checked because the markdown checks in this file expect a recipe file.

danielleontiev: Imagine you have a list of files in the diff, f.e: `foo.md`, `.github/baz.md`, `bar.webp`. Script will exit with 0 exit code after it meets the second file and `bar.webp` will not be checked

.github/workflows/scripts/check-files.sh
hallzy: Fair enough. In fact, I think the way it is written now doesn't work if the file is in the wrong folder either, as it just prepends the folder to the filename without fixing the folder it is in.

So my latest update actually just removes that regex altogether

.github/workflows/scripts/check-files.sh
hallzy: woops. Thanks for catching that. This has been merged now though, so I'll have to open up a new PR

danielleontiev:
```sh
#!/bin/sh

echo "foo.md\n.github/baz.md\nbar.webp" | while IFS= read -r file; do
    case "$file" in
        *.webp)
            echo "WEBP file $file"
            ;;
        .github/*.md)
            echo "MD in .github $file"
            exit 0
            ;;
        *.md)
            echo "MD to check $file"
            ;;
    esac
done
```

Take a look at this example. Try to run it and then delete `exit 0` part and run again

@LukeSmithxyz you are hurring to much

PR #402, opened 2021-03-22
Fixed some issues from last PR #400

hallzy:
Made some mistakes that @danielleontiev found in my last PR. This is to fix those.

I noticed that actually at any point if something fails that it wouldn't check the rest of the files, so I removed the exit commands and we now exit once at the end of all the files depending on if we had any failures.


danielleontiev (approved):
Thank you for fixing it!

PR #419, opened 2021-03-22
removed commas in tag section

lxgr-linux:
removed commas in tag section for glutenfree-pfannkuchen


hallzy:
This change is already in an open PR #392

PR #423, opened 2021-03-23
added updated lenten chili recipe with picture

yorgez:


src/lenten-chili.md
```diff
@@ -0,0 +1,41 @@
+# Lenten (Vegan) Chili
+
```
Knarkzel, line 2: Missing image: `Lenten chili`

src/lenten-chili.md
```diff
@@ -0,0 +1,41 @@
...
+6. Add each of the spices, taste the chili, and adjust flavoring accordingly.
+7. Let cook for 20-30 minutes, stirring occasionally.
+
+Let cool for a few minutes and then serve in bowls.
+
+## Contribution
+
+- George Z
+
+;tags: chili, vegan, lenten
```
Knarkzel, line 41: Tags should be written like so: `;tags: chili vegan lenten`, space seperated.

hallzy: fyi, the recipe checking script should catch this from now on

PR #424, opened 2021-03-23
better wording

peepee:
using "via" twice in the same breath is strange


hallzy (approved):

hallzy:
Just an FYI I've made a PR #426 so that these pull request checks don't fail if index.md is changed.

No changes needed here for that, just wanted to point out that the failed check here has been addressed.

Knarkzel (approved):

PR #428, opened 2021-03-23
Add picture to Cinque Pi

pahlse:


Knarkzel:
You're missing a link to the image like so: `<Recipe title>`.

PR #441, opened 2021-03-23
added image for bolognese sauce

yiusa:


DGKSK8LIFE (approved):

src/bolognese-sauce.md
```diff
@@ -1,5 +1,7 @@
 # Bolognese Sauce
 
+![Bolognese sauce](pix/bolognese-sauce-1.webp)
```
Knarkzel, line 3: Are there multiple images? If not, the image should be called `bolognese-sauce.webp` instead.

hallzy (approved):

PR #445, opened 2021-03-23
Fix spacing in recipes from old example

Morelcia:
suicide fuel


Knarkzel (approved):

hallzy (approved):
An optional change you could make is to fix all the issues in the check script on this PR. None of the issues found are a result of what you did, but were pre-existing, so feel free to fix those if you want.

Just click on 'details' on the check and you should be able to find a list of files that were checked, and any problems found in them.

LukeSmithxyz:
Yeah it didn't appear obvious on my machine due to how my browser displays the emoji so I didn't notice.

PR #447, opened 2021-03-23
Make few recipes compliant with submission rules

Morelcia:


Morelcia:
Is there any easy way to get list of every broken file? 
Just running `.github/workflows/scripts/check-files.sh` doesn't work and I'm too retarded to modify it

hallzy:
> Is there any easy way to get list of every broken file?
> Just running `.github/workflows/scripts/check-files.sh` doesn't work and I'm too retarded to modify it

I if you change:

```
$(git diff --name-only "$(git merge-base origin/master HEAD)")
```

to:

```
$(git ls-files)
```

That would probably work...

although, I think the script will actually run on files like example.md and README.md still, so ignore those results. We should probably add exceptions for those files to the check-files.sh script as well

src/garlic-toast.md
```diff
@@ -1,28 +1,28 @@
...
-5. Bake for 10-15 minutes or until crunchy throughout.
-6. Serve with cheese, charcuterie, pickles, dips, or to accommodate a soup. 
-
-## Contribution
-
-- Jack Crawford - [website](https://github.com/jjcrawford)
-- Edited by Zyansheep
-
-;tags: basic bread quick side
+# Garlic Toast
```
hallzy, line 1: What changed in this file? not really sure what happened with this diff

Morelcia: dos->unix

hallzy: k good to know. Don't think it says anywhere

hallzy (approved):

PR #448, opened 2021-03-23
Create southern-biscuits.md

pard68:
There is salt in the list, but I believe this would be an exception, as in baked goods you need a precise amount of each ingredient and not just an eyeballed or to-taste amount.

### DIFF
--- a/src/southern-biscuits.md
+++ b/src/southern-biscuits.md
@@ -1,0 +1,32 @@
+# Southern Biscuits
+
+This is Southern (American) Style Biscuits.
+
+## Ingredients
+
+- 2 cups AP flour
+- 1 tsp white sugar
+- 1 tbsp baking powder
+- 1 tsp kosher salt
+- 8 tbsp sweet butter
+- 3/4 cup whole milk
+
+## Directions
+
+1. Preheat oven to 425°F
+2. In a bowl combine dry ingredients
+3. Cube butter and add to dry ingredients
+4. Mix butter with hands or fork until a cornmeal consistency is achieved
+5. Add milk and mix until homogeneous
+6. Roll biscuits out to approx. 3/4 of an inch thick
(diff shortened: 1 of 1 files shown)


hallzy (approved):
any chance you have any images to add?

pard68:
Probably won't have an image to add for a week or so. Usually biscuits are a Wednesday thing, but we're moving Thursday so everything is all packed up right now.

hallzy:
> There is salt in the list, but I believe this would be an exception, as in baked goods you need a precise amount of each ingredient and not just an eyeballed or to-taste amount.

Just came across another baking recipe someone submitted with salt in the ingredients. I think it if it is baked in then making an exception is fine, so long as you have given a measurement (like you have).

I think the intent of that rule in the first place was to avoid having every recipe have it because you add it for taste the end of basically everything.

This is just a long winded explanation to say that I have no problem with the salt in the ingredients list in this case.

src/southern-biscuits.md
```diff
@@ -0,0 +1,32 @@
...
+2. In a bowl combine dry ingredients
+3. Cube butter and add to dry ingredients
+4. Mix butter with hands or fork until a cornmeal consistency is achieved
+5. Add milk and mix until homogeneous
+6. Roll biscuits out to approx. 3/4 of an inch thick
+7. Use a mug, cup, or cookie cutter to cut out rounds of dough
+8. Arrange rounds in a cast iron pan
+9. Bake for 12 to 15 minutes
+10. Remove from oven and eat hot
+11. Will keep on counter or in a brisker
```
Socerest2, line 26:
```suggestion
11. Keep them on counter or in a brisker
```
Could be my English but I think this would be a better way to say this.

Socerest2 (changes requested):

PR #451, opened 2021-03-23
Fixes

Morelcia:
Everything should be now compliant with rules except from webp image sizes


Socerest2:
Remember to compress your files!!!! arroz-chaufa-2 was too big. I compressed it for you.
[arroz-chaufa-2](https://i.ibb.co/PxLN5s0/arroz-chaufa-2.webp)

Morelcia:
nice meme

Socerest2:
> nice meme

huh, thats odd, it should off passed. You uploaded the file, it's under 100kb. It's erroring saying it can't find the file.

Morelcia:
Should I git reset this compression and force push?

Socerest2:
> nice meme

Found it, rename my file i gave you arroz_chaufa-2. I accidentally labeled it with a `-`

Morelcia:
It's supposed to be with `-`

LukeSmithxyz:
I don't know what you're doing with tags, but none of those make sense.

No one is going to be looking for "salt" or "pepper" or "onion" or "meat". They're not supposed to be ingredients. The tag cloud is bigger than it should be already. Keep to the meat used, the region of origin and large themes for tags.

src/arroz-chaufa.md
```diff
@@ -2,7 +2,7 @@
 
 Peruvian-chinese dish. Easy to cook just add and mix everything.
 
-![Arroz Chaufa](pix/arroz_chaufa-2.webp)
+![Arroz Chaufa](pix/arroz-chaufa-2.webp)
```
Socerest2, line 5:
```suggestion
![Arroz Chaufa](data/pix/arroz-chaufa-2.webp)
```
Fixed Error

src/arroz-chaufa.md
```diff
@@ -18,7 +18,7 @@ Peruvian-chinese dish. Easy to cook just add and mix everything.
 - Soy Sauce
 - Welsh Onion
 
-![Ingredients](pix/arroz_chaufa-1.webp)
+![Ingredients](pix/arroz-chaufa-1.webp)
```
Socerest2, line 21:
```suggestion
![Ingredients](data/pix/arroz-chaufa-1.webp)
```
Fixed Error

Socerest2 (changes requested):
Found the check error!

Morelcia:
> I don't know what you're doing with tags, but none of those make sense.
> 
> No one is going to be looking for "salt" or "pepper" or "onion" or "meat". They're not supposed to be ingredients. The tag cloud is bigger than it should be already. Keep to the meat used, the region of origin and large themes for tags.

I'll fix tags, should I keep "meat" tag? It's already in `brown-sauce` anyway

src/arroz-chaufa.md
hallzy: It is correct without the `data`. The webserver file structure is slightly different to the repository, so you don't need the `data` folder. Yes, it looks bad when you visit the markdown file here on github, but that's how it is

src/arroz-chaufa.md
hallzy: Same as above. `data` should not be there.

Morelcia:
Should be better now. I tried to limit making new tags but it's just impossible to get 2 tags in something as simple as corn salsa

hallzy:
Looks like the checks are failing on this file `data/pix/arroz_chaufa-2.webp` cause it doesn't exist... Seems like an oversight in the script. I guess because you renamed it, but the renamed file is now different because it is smaller...

We should really have a way to not check files that are deleted in the script

src/arroz-chaufa.md
Socerest2: > It is correct without the `data`. The webserver file structure is slightly different to the repository, so you don't need the `data` folder. Yes, it looks bad when you visit the markdown file here on github, but that's how it is

thanks!

Socerest2 (approved):

Morelcia:
> Looks like the checks are failing on this file `data/pix/arroz_chaufa-2.webp` cause it doesn't exist... Seems like an oversight in the script. I guess because you renamed it, but the renamed file is now different because it is smaller...
> 
> We should really have a way to not check files that are deleted in the script

It was working correctly until I replaced old file with compressed one.
Also if tags are still bad then this task is not for my brain

hallzy (approved):

LukeSmithxyz:
Get rid of meat, corn, salsa and eggs as tags. If you want corn salsa to have appropriate tags, sauce or mexican would be the obvious ones. Yeah brown sauce has/had meat but it'll be gone in a bit.

Aethusx (approved):

Morelcia:
meat, corn, salsa and eggs are gone

PR #456, opened 2021-03-23
add fresh lasagna recipe and picture

yiusa:


LukeSmithxyz:
There's already a lasagna recipe.

yiusa:
> There's already a lasagna recipe.

oh yeah was supposed to change that one, if you take a look at it i find it quite lackluster since it uses store bought pasta sauce etc.

LukeSmithxyz:
Make the changes to his article and then link to separate directions for sauce so people can either follow those extra directions or buy their own. Looking on what you have, it looks like you sort of already have that in this article. You can feel free to mostly replace his.

Also, use proper markdown links. If you just paste the URL, it won't link properly.

yiusa:
will do!

PR #457, opened 2021-03-23
Minor ReadMe Edit

Socerest2:
Adds instruction when img is too big.

### DIFF
--- a/README.md
+++ b/README.md
@@ -10,7 +10,7 @@ There are no ads, trackers, cookies (unless recipes thereof) or javascript.
 - By adding a recipe.
 - Make a recipe and take a nice picture of it if no nice picture already
   exists. Submitted images should be small `.webp` files ideally less than 100K
-  or so. If your image is above 100K, shrink the size [here](https://www.compresss.com/compress-webp.html#), since if it's bigger then the limit, the check won't pass.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -10,7 +10,7 @@ There are no ads, trackers, cookies (unless recipes thereof) or javascript.
 - By adding a recipe.
 - Make a recipe and take a nice picture of it if no nice picture already
   exists. Submitted images should be small `.webp` files ideally less than 100K
-  or so.
+  or so. If your image is above 100K, shrink the size [here](https://www.compresss.com/compress-webp.html#), since if it's bigger then the limit, the check won't pass.
```
hallzy, line 13:
```suggestion
  or so. If your image is above 150K, shrink the size [here](https://www.compresss.com/compress-webp.html#), since if it's bigger then the limit, the check won't pass.
```

The check actually fails on fails over 150kB.

hallzy (changes requested):

hallzy:
Ignore the failed check. It is checking the README as if it were a recipe.

Socerest2:
> Ignore the failed check. It is checking the README as if it were a recipe.

Yep, and I fixed the 100 to 150

hallzy (approved):

PR #462, opened 2021-03-24
More checks for PRs

hallzy:
- PR script now ignores example.md and README.md
- PR script attempts to validate image inclusions
    - does the image exist?
    - is the markdown syntax valid for showing an image?
- Checks to make sure a file exists before checking it (this is for the case where a file is deleted)
- Checks for trailing whitespace
- Checks for tabs that are anywhere in the document except for the beginning as indentation
- Checks for CRLF line endings
- More descriptive messages
- Only prints out information about files that have problems instead of polluting the output by announcing that we are checking every single file.
- Check to make sure all files have an extension (the only way to really do that in a case statement is to check for `*.*`, which is probably good enough anyways).

The awk script was getting long, so I split it off into it's own script

### DIFF
--- a/.github/workflows/scripts/check-files.sh
+++ b/.github/workflows/scripts/check-files.sh
@@ -61,6 +61,8 @@ check_recipe() {
 }
 
 while IFS= read -r file; do
+	errMsgs=''
+
 	# If the file doesn't exist, then the user probably deleted it, so don't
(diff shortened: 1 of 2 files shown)


zyansheep (approved):
gawk is pretty cool...

PR #469, opened 2021-03-24
Update and rename lasagna.md to lasagne.md

underpayed-driller:
In italian this dish is called 'lasagne', not 'lasagna', translated to English it's still 'lasagne' (https://www.deepl.com/translator#it/en/lasagne).
I fixed this because it increases the integrity of the site and makes it look more serious.

### DIFF
--- a/src/lasagne.md
+++ b/src/lasagne.md
@@ -1,15 +1,16 @@
-# Lasagna
+# Lasagne
 
-A classic italian dish, lasagna is a great meal for a cold day when you're in the mood for pasta. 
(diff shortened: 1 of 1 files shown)


LukeSmithxyz:
Don't know if you noticed, but the site is in English.

underpayed-driller:
> Don't know if you noticed, but the site is in English.

I did, in fact, notice that.
That's exactly my point. In English it's still lasagne, eventhough it is spelled with an 'a' very often. Nevertheless, I don't want to annoy you, have a great day.

PR #472, opened 2021-03-24
Python Utility Script - Tag Editor

DollarStarNova:
image

This python script allows easy mass-auditing and editing of tags on recipes to allow easy addition of things like vegan, vegetarian, gluten-free that apply to a wide range of recipes but recipe authors may not have considered adding as tags.

Note: I am aware there are other specified tags for non-meat non-dairy but these are not normal terms and if we want this site to be accessible we should use normal terms like vegetarian and vegan. Additionally, vegetarian / vegan mean more than just meat / dairy free, so they are more accurate terms.

### Usage:

`python tag-editor.py`

Then for every recipe in the src folder, you will be presented with a prompt showing current recipe, ingredient list and current tags.

you can add tags, remove tags, go to the next item or quit:

### Adding Tags 

Command: `[add / a / +] {tag tag tag...}`

Example: 

`add chicken gluten-free`   --- > Adds "chicken" and "gluten-free" tags to current recipe

### Removing Tags 

Command: `[remove / r / -] {tag tag tag...}`

Example: 

`remove pork beef mexican`   --- > Removes "pork", "beef" and "mexican" tags from current recipe (if present)

### Next Recipe

Command: `next/ n / [Return_Key]`

Moves to next recipe in the folder.

### Quit

Command: `quit / q` --> quits the program.

Program will also terminate when it has cycled through all recipes in the list.


DGKSK8LIFE:
Could you document usage of it in the thread? Thanks @DollarStarNova

DollarStarNova:
> Could you document usage of it in the thread? Thanks @DollarStarNova

I have now documented this in the top comment :)

DGKSK8LIFE:
Awesome @DollarStarNova

DGKSK8LIFE (approved):

DollarStarNova:
@LukeSmithxyz Any reason this has been rejected? I think it's a very useful tool.

DGKSK8LIFE:
Agreed @DollarStarNova

PR #477, opened 2021-03-24
added "dynamic" description to all pages and fixed build error

canicjusz:
Now it will take the first line which isn't a title, an image or a list element and set it as content in the description meta tag.
I also added default description for the main page and tag pages.

I removed an apostrophe from some cooking recipe's name, which made website not to build.


hallzy:
#475 has the same fix to the build that you have. I also changed the name of the image because it had the same problem (the makefile probably doesn't do much with the pictures though so it probably passes without it, but I did the change just for consistency)

canicjusz:
Ok, I just changed the picture's name too.

PR #486, opened 2021-03-25
Added Basic Meatball Recipe

jhubberts:
Included a basic meatball recipe I modified from some others I found on blogs and in J Kenji Lopez Alt's "Food Lab".

### DIFF
--- a/src/basic-meatballs.md
+++ b/src/basic-meatballs.md
@@ -19,18 +19,19 @@ Hybrid beef/pork meatballs perfect for spaghetti and meatballs or a meatball sub
 - 1 tbsp rosemary
 - 1 tbsp basil
 - 1 tbsp garlic powder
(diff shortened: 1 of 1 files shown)


hallzy (approved):
Your directions are all numbered with a `1`. I don't think it matters too much because markdown usually numbers things for you anyways, but thought I'd point it out if you want to fix it.

jhubberts:
> Your directions are all numbered with a `1`. I don't think it matters too much because markdown usually numbers things for you anyways, but thought I'd point it out if you want to fix it.

Yep that was by intention, I use Markdown for READMEs at work and at least one guy I work with insists that you should always use 1. for numbered lists and rely on Markdown to format it correctly in case you have to insert a new step in the middle. Kind of a pain but I'm used to it at this point and also have come around to his opinion on the matter.

PR #487, opened 2021-03-25
gotta go fast: don't clean before building

sylGauthier:
Initially, I made blogit to clean everything before rebuilding because the Makefile was still being modified (and rebuild is not triggered by a Maekfile modif). NMow, with the growing number of recipes, it's time to use the Makefile in all its glory and let it decide for itself what to rebuild.

### DIFF
--- a/.github/workflows/upload.yml
+++ b/.github/workflows/upload.yml
@@ -30,4 +30,4 @@ jobs:
           cd repo
           git stash
           git pull --force origin master
-          make clean deploy
+          make deploy


.github/workflows/upload.yml
```diff
@@ -30,4 +30,4 @@ jobs:
           cd repo
           git stash
           git pull --force origin master
-          make clean deploy
+          make deploy
```
DGKSK8LIFE, line 33: Not sure about this

hallzy: Any particular reason?

DGKSK8LIFE: Well even if `make clean deploy` adds more latency, I think it should still auto-clean everything. But I'm ok with it deciding what to clean as well.

sylGauthier: That's really how blogit is meant to be used tho. `clean` here means all previous recipe pages are destroyed, but the Makefile is relatively smart, it knows exactly which one to rebuild, which tag page to rebuild, etc. The only reason why I put `clean` in the first place was because I was still heavily modifying the Makefile, and when the Makefile is changed, you do need to rebuild everything (since the Makefile is not tracking itself as a dependency).

DGKSK8LIFE: Ok. Then I'm in agreeance with this change. Good explanation. Thanks.

DGKSK8LIFE (approved):

hallzy (approved):

LukeSmithxyz:
Won't this be leaving up orphaned pages and tags?

sylGauthier:
Here is what I can tell from my tests:
- Recipe added: builds only new recipe.
- Recipe removed: requires full rebuild (or at least index, a solution to mitigate this would be to clean/rebuild the index every time).
- Recipe modified: only rebuild recipe.
- Tag added to an existing recipe: all tag pages containing the recipe, the index and the recipe get rebuild, which is the expected behavior.
- Tag removed from existing recipe: same as above.

Now if you remove completely a tag, its tag page will indeed remain but won't be linked anywhere. It would be relatively easy to make a rule to remove them, and ask rsync to remove files in the dest that aren't in the source.

PR #513, opened 2021-03-28
Add Teriyaki-sauce-light

modusmoe:

### DIFF
--- a/src/Teriyaki-sauce-light
+++ b/src/Teriyaki-sauce-light
@@ -1,0 +1,27 @@
+# Teriyaki sauce (light)
+
+A light teriyaki sauce which keeps in the cupboard for months.
+
+- ⏲️ Prep time: 4 min
+- 🍳 Cook time: 5 min
+- 🍽️ Servings: 20
+
+## Ingredients
+
+- 200 ml mirin.
+- 200 ml sake
+- 200 ml soy sauce (optional: increase to 250 ml or to taste for darker teriyaki)
+- 30 g sugar (optional: increase to 60 g or to taste for sweeter, thicker glaze)
+
+## Directions
+
+1. put all ingredients in a small saucepan
+2. stir over low heat until sugar is fully dissolved (do not boil or reduce)
+3. pour into an empty wine bottle or similar
+
+## Contribution
+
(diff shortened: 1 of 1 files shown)


src/Teriyaki-sauce-light
```diff
@@ -0,0 +1,27 @@
...
+1. put all ingredients in a small saucepan
+2. stir over low heat until sugar is fully dissolved (do not boil or reduce)
+3. pour into an empty wine bottle or similar
+
+## Contribution
+
+- moe@modus.uno - Donate Bitcoin: `bc1q4ux9csqeh55kmagu5vyl6hjejqfpql6kv730yz` - Donate Monero: `49uVPc36BdAMH3uptGPn5U9JJ42hoCptBMfndegAj4Pi3Kfu7pUSFgMgc5oSwdzrXMgX3vpAZ6YbpgN2D5TX1bUVDofS7oN`
+
+;tags: Japanese Teriyaki Sauce Quick
+\n
```
hallzy, line 27:
```suggestion
;tags: japanese sauce quick basic
```

Please remove the `\n` line, and lowercase all tags. Also can you remove the teriyaki tag. I think the other 3 are good enough, and we don't currently have a teriyaki tag and the list of tags is already quite long.

I think the basic tag would go well here as well. The `basic` tag is reserved for things that aren't meant to be standalone and are to be used in other dishes.

hallzy (changes requested):
Can you please rename the recipe file to `src/teriyaki-sauce-light.md`. ie, all lowercase and with the `.md` file extension

src/Teriyaki-sauce-light
modusmoe: Done, thank you for the input, will keep in mind next time.

hallzy (approved):

src/Teriyaki-sauce-light
modusmoe: This is my first github commit.  What d i need to do to finalise and for the recipe to be  published?

hallzy: Nothing on your end, just need to wait for Luke to merge it now.

PR #515, opened 2021-03-28
make footer navigatable with #s in address bar

Odyssey346:
This makes it able to navigate the index site with just putting a #tag in the address bar.

So, you can do, for example:
`https://based.cooking/#about` and it'll show you the about 

oh yeah this also meant I had to convert index.md to html, and I put it in an unused file so no space was wasted :P 

works on my machne


hallzy:
From the documentation on `discount` you can add the `toc` and `idanchor` flags to do this automatically for everything. This would make this work for all recipes as well.

ex:

```
markdown -f idanchor,toc -s "
# This is a test

## Sub heading

This is some text
"
```

Produces this HTML:

```
<h1 id="This-is-a-test">This is a test</h1>

<h2 id="Sub-heading">Sub heading</h2>

<p>This is some text</p>
```

Separate issue is, I don't know if this will be merged tbh. I find that this is most useful when a page is long, but there aren't really any long pages on this website. But, we can see I guess.

Odyssey346:
> From the documentation on `discount` you can add the `toc` and `idanchor` flags to do this automatically for everything. This would make this work for all recipes as well.
> 
> ex:
> 
> ```
> markdown -f idanchor,toc -s "
> # This is a test
> 
> ## Sub heading
> 
> This is some text
> "
> ```
> 
> Produces this HTML:
> 
> ```
> <h1 id="This-is-a-test">This is a test</h1>
> 
> <h2 id="Sub-heading">Sub heading</h2>
> 
> <p>This is some text</p>
> ```
> 
> Separate issue is, I don't know if this will be merged tbh. I find that this is most useful when a page is long, but there aren't really any long pages on this website. But, we can see I guess.
> 
> A fun extra, if you add the `-T` option, discount will actually create a table of contents for you with links to these sections. Don't think we need that though

Can we manually define them?

hallzy:
Don't think so. Looks like the IDs just come from whatever the content of the tag is.

Odyssey346:
wtaf i dont remember making this PRSendt fra min Galaxy
-------- Opprinnelig melding --------Fra: Luke Smith ***@***.***> Dato: 29.06.2022  16:27  (GMT+01:00) Til: "LukeSmithxyz/based.cooking" ***@***.***> Ko: Odyssey346 ***@***.***>, Author ***@***.***> Emne: Re: [LukeSmithxyz/based.cooking] make footer navigatable with #s in address bar (#515) 
Closed #515.

PR #520, opened 2021-03-29
added spicy mayo recipe

scary900:


src/spicy-mayo.md
```diff
@@ -0,0 +1,27 @@
...
+
+## Note
+
+This is just a basic guide, adapt this recipe to your liking.
+
+## Contribution
+
+- scary90
+
+;tags: simple sauce japanese
```
hallzy, line 27:
```suggestion
;tags: basic sauce japanese
```

Just to avoid adding another tag to the long list, I think using `basic` instead of simple would be best. I think this would also fit in with the basic category

hallzy (changes requested):

hallzy (approved):

PR #532, opened 2021-03-31
Russian bliny recipe

Yaroslav-95:


DGKSK8LIFE (approved):

src/russian-bliny.md
```diff
@@ -0,0 +1,52 @@
...
+
+* 300—400g wheat flour
+* 700—900ml milk
+* 2 teaspoons baking soda
+* 2—3 eggs
+* 1 teaspoon olive oil or melted butter
+* Butter
+* Boiling water
+* Sugar
+* Salt
```
hallzy, line 22:
```suggestion
```

Don't need salt in ingredients list

src/russian-bliny.md
```diff
@@ -0,0 +1,52 @@
...
+
+## Directions
+
+1. You should start by adding the flour to a bowl, then add the eggs, sugar and
+   salt to taste, and baking soda.
+2. Next add the milk carefully, so that the batter ends up liquid enough, but
+   not too liquid. It should be more liquid than bread batter, but not too
+   liquid so that it doesn't drip too fast.
+3. Add a little bit of boiling water and oil or butter and mix the batter
+   until it becomes homogeneous. If you feel its too liquid, add a little bit of
```
hallzy, line 32:
```suggestion
   until it becomes homogeneous. If you feel it's too liquid, add a little bit of
```

src/russian-bliny.md
```diff
@@ -0,0 +1,52 @@
...
+   flour, if it's not liquid enough, then add a little bit more milk.
+4. Heat up a frying pan to medium heat, cover with a thin layer of melted
+   butter, you can use a paper napkin or towel for that.
+5. Slowly pour the batter to the pan rotating it a little bit while you do this,
```
hallzy, line 36:
```suggestion
5. Slowly pour the batter into the pan rotating it a little bit while you do this,
```

src/russian-bliny.md
Yaroslav-95: I would argue that it is not always obvious if a recipe needs salt/pepper or not, but I guess the rules are the rules ¯\_(ツ)_/¯.

hallzy: I would agree. But it is mentioned in the directions and that is probably enough. Leaving it in the directions so there is an obvious place to add it is fine.

hallzy (approved):

m-chrzan:
I would simplify name to just "bliny" (or "blini"? looks like Wikipedia prefers that englishization).

m-chrzan:
Also, looks like #498 also has a bliny recipe. Ideally we merge only one (potentially taking best parts of both).

Yaroslav-95:
> I would simplify name to just "bliny"

I think it's a matter of style or preference, not to mention there might be Ukranian or Belarussian bliny. After all, there's a "French crepes" recipe on based.cooking already and nobody seemed to bat an eye on that.

>(or "blini"? looks like Wikipedia prefers that englishization).

Once more, a matter of style, even Wikipedia mentions this transliteration (bliny), not to mention that Wikipedia is not the best source of information. The logic behind this transliteration is that that the i and y represent different sounds (и and ы) so from my point of view as a Russian speaker "bliny" makes a lot more sense than "blini". блины → bliny.

> Also, looks like #498 also has a bliny recipe. Ideally we merge only one (potentially taking best parts of both).

Hmm, hadn't noticed that one. I also guess it would make sense to keep only one, although both instructions differ a bit. Maybe keep one of these, but mention both as contributors?

Issue #150, opened 2021-03-13
Search?


marekkowalczyk:
Let's have a search box, in due time.

Morelcia:
You can just use ctrl+f in browser.
I guess some simple php for searching could be useful if in the future categories on separate pages would be introduced.

Goblin80:
#153

Godje:
@Morelcia I assume the point is to not have a serverside preprocessor. Currently the side is literally assembled through a shell script. A search box with a JS filter function that has an array of pages would do just fine, and would be serverside. It simple enough and light enough to not need to be cached (which means we could just put it in the head of the script).

Morelcia:
Afaik point is to be as lightweight as possible clientside and to work without js.
+serverside js abominations definitely aren't lightweight unless you meant something else.

Godje:
@Morelcia definitely not serverside. Client side small amount of javascript (6 to 10 loc), inside the <head> `<script>` tag.

Godje:
Since the amount of script I want to add is miniscule, it has no need to exist in a separate Network call to import the Javascript. So we can put it in the `<head>` tag, or at the bottom of HTML and forget it.
When you don't import the stuff in a separate file, it doesn't get cached. But since the amount of code is so small, there is no need for caching.

Goblin80:
@Godje I see you point, and if you ask me i have no problem with minimal JS. but i don't think luke and the community will approve that. he was pretty clear about a no JS policy.

marekkowalczyk:
How about just an alphabetized static index page of user-marked [[keywords]]?

adeptflax:
How about JavaScript on the server-side? [http://www.cgi-node.org/](http://www.cgi-node.org/)

Morelcia:
Why would someone run backend in language which runtime takes 500MB ram in idle? unironically php>nodejs

adeptflax:
@Morelcia lol no it doesn't

adeptflax:
ok serious this time, the more efficient thing would be a c++ nginx plugin.

oussamabouchareb:
ctrl+f should be more than enough

adeptflax:
the laziest way would probably be bash-cgi script.

Dko1905:
> ctrl+f should be more than enough

The chad way to search is to use the inbuilt search.

adeptflax:
actually making a duckduckgo search form would be even easier

adeptflax:
```html
[...]

PR #533, opened 2021-04-01
Add search feature

omar2205:
Adds search feature to the home page
image

### DIFF
--- a/data/authors/benjamin-h.json
+++ b/data/authors/benjamin-h.json
@@ -1,5 +1,5 @@
 {
 	"name": "Benjamin H",
 	"website": "https://beherrlich.xyz",
-	"XMR": "yTQ2hifnHNnqHNMQNqE6nggMGCSqsXadThkrFhTWKo4dK6HFj6NDUocnJYHvaQjZ1TwLDMYXVpj6bCm2Sonhpnp"
+	"xmr": "yTQ2hifnHNnqHNMQNqE6nggMGCSqsXadThkrFhTWKo4dK6HFj6NDUocnJYHvaQjZ1TwLDMYXVpj6bCm2Sonhpnp"
 }


omar2205:
The smallest amount of JS to search from recipes list

hallzy:
Short of not having the variables, this is probably the shortest way to do it, and doesn't require continuously removing and adding DOM elements.

```javascript
const rec = document.querySelectorAll('#artlist li')
const search = document.querySelector('#search')

search.addEventListener('change', e => {
    const searchText = e.target.value.toLowerCase();

    rec.forEach(el => {
        const recipeName = el.innerText.toLowerCase();
        const isMatch = recipeName.indexOf(searchText) > -1;

        el.hidden = !isMatch;
    });
});
```

That being said, I will be shocked if this gets into the project.

This search is no different to a standard ctrl+f search, and every issue I have seen created about adding any javascript at all has been closed.

one ref: #150

omar2205:
yes there is the CTRL+F, but to type & search is just easier

omar2205:
image

LukeSmithxyz:
I am open to this. I will test it out when I can.

LukeSmithxyz:
To be clear, yes, search via ctrl-f is possible, but this avoids the issue of visual overwhelm. Users can see all "pasta" entries in one place, rather than ctrl-fing time after time again to different parts of the doc.

Important to me is it maintaining perfect compatibility with cli access.

DGKSK8LIFE:
So you are ok with contributing Javascript to this repo now? @LukeSmithxyz Hmm .

hallzy (changes requested):
If this is something we are going to add, then this is my review. The search does not work for me.

The script is executed before the recipe list is rendered, so the script doesn't know about any recipes.

We will either have:

- Move this into the footer file
- wrap the script in an onload or something, or
- take the script out and put it into a separate `.js` file, and reference it with `<script defer ...>` in this file in the head.

templates/article_list_header.html
```diff
@@ -1,2 +1,31 @@
...
+    const isMatch = recipeName.includes(searchText)
+
+    el.hidden = !isMatch
+  })
+})
+  
+clearSearch.addEventListener('click', e => {
+  search.value = ''
+  rec.forEach(el => el.hidden = false)
+})
```
hallzy, line 29:
```suggestion
window.addEventListener('load', () => {
  const rec = document.querySelectorAll('#artlist li');
  const search = document.querySelector('#search');
  const clearSearch = document.querySelector('.clear-search');

  search.addEventListener('change', e => {
    // grab search input value
    const searchText = e.target.value.toLowerCase();
    // for each receipe hide all but matched
    rec.forEach(el => {
      const recipeName = el.innerText.toLowerCase();
      const isMatch = recipeName.includes(searchText);

      el.hidden = !isMatch;
    });
  });

  clearSearch.addEventListener('click', e => {
    search.value = '';
    rec.forEach(el => el.hidden = false);
  });
});
```

This for example, would work.

hallzy (changes requested):

hallzy (approved):

PR #536, opened 2021-04-02
Spaghetti and Meatballs photo

ClosedGL2:


src/spaghetti-and-meatballs.md
```diff
@@ -82,6 +84,6 @@
 
 ## Contribution
 
-- ClosedGL - [GitHub](https://github.com/ClosedGL2)
+- ClosedGL -- [GitHub](https://github.com/ClosedGL2), [PeerTube](https://truetube.media/https://truetube.media/video-channels/cgl/videos)
```
hallzy, line 87:
```suggestion
- ClosedGL -- [GitHub](https://github.com/ClosedGL2), [PeerTube](https://truetube.media/video-channels/cgl/videos)
```

Looks like you have `truetube.media` twice in the url

hallzy (changes requested):
Could you also reduce the size of the image. It is 192kB and should be less than 150kB

src/spaghetti-and-meatballs.md
ClosedGL2: Fixed

PR #538, opened 2021-04-02
Added garam masala recipe

chetanbasuray:


src/garam-masala.md
```diff
@@ -0,0 +1,38 @@
...
+- 4 Green Cardamom
+- 2 Black Cardamom
+- 1 tbsp Coriander
+- 2 Cloves
+- 1 tbsp Cumin
+- 1 tbsp Fennel seeds
+- 5/6 Bayleaf
+- 5/6 Black pepper pods
+- 1 Star anise (Optional)
+- 0.25 Mace (Optional)
```
hallzy, line 16: 0.25 of what unit?

Also, maybe I am just ignorant but I've never heard of mace before. Thought for a second we were talking about pepper because mace is a brand of pepper spray.

If it is common and I am just an idiot, ignore me. But otherwise a short description of what it is may be useful. Looks like it comes from the nutmeg seed, so maybe nutmeg could be mentioned as well.

src/garam-masala.md
```diff
@@ -0,0 +1,38 @@
...
+
+## Directions
+
+1. Star anise and Mace are normally not added to Garam masala. They are added specially if someone is making [Biriyani](lamb-biriyani.html).
+2. Cloves will make the masala really hot. So make sure to add only as much as you want to eat.
+3. Too much Mace makes the masala quite bitter. So make sure you donot add more than what has been recommended. The amount of other spices can be changed according to taste.
```
hallzy, line 22:
```suggestion
3. Too much Mace makes the masala quite bitter. So make sure you do not add more than what has been recommended. The amount of other spices can be changed according to taste.
```

src/garam-masala.md
```diff
@@ -0,0 +1,38 @@
...
+4. Wash all the dry spices really well.
+5. If you live in a place which is very sunny, sun dry your washed spices for about 1 day. If not then on a frying pan with a heavy bottom, place all the spices and dry roast them until aromatic.
+6. Cool the dry spices and put them into a mixer. Mix thoroughly.
+7. Sieve the powder and add the larger pieces into the mixer once again and mix it again till everything turns into a fine powder. This step might need to be repeated a few times.
```
hallzy, line 26:
```suggestion
7. Sieve the powder and add the larger pieces into the mixer once again and mix it again until everything turns into a fine powder. This step might need to be repeated a few times.
```

hallzy (changes requested):

src/garam-masala.md
chetanbasuray: Yes, Mace is created from the Nutmeg seed and it generally comes in a fixed size. I meant 0.25 of that one piece of Mace. Not sure what would be the best way to write this.

hallzy (approved):

PR #541, opened 2021-04-03
Merge similar pancake recipes

DavGoeck:


src/pancake.md
```diff
@@ -1,25 +1,33 @@
...
+- Syrup
+- Cinnamon and sugar
 
-BeFe
+## Contribution
+- BeFe (original)
+- Lxgr (gluten-free) - [github](https://github.com/lxgr-linux)
 
-;tags: quick breakfast sweet pancake cheesefare
+;tags: quick basic breakfast sweet pancake cheesefare
```
hallzy, line 33:
```suggestion
;tags: quick breakfast sweet pancake cheesefare
```
Not sure if basic should be added. It's really meant for recipes that are not meant to be done alone as a dish, but pancakes on their own (except for the addition of syrup or something) can pretty much be their own thing.

hallzy (approved):

src/pancake.md
hallzy: Also, 6 tags is too many

hallzy:
Failed check can be ignored. The newer version of the PR script I have in a PR (#462) wouldn't fail. it is just because a file was removed in this PR and the current PR script doesn't account for that.

DGKSK8LIFE (approved):

DavGoeck:
I think that all pancake recipes should be merged into a single one. The differences are marginal and can be recorded as variations similar to the egg recipe. The only thing holding me back from doing it is that I never made pancakes and cannot judge which one is the most standard one. I also don't want to screw up the recipes when rescaling the different recipes to a common denominator.

siggines:
This should not be pulled, it is too similiar to https://based.cooking/pancake

PR #545, opened 2021-04-03
Added lentil curry recipe

xinnna:


hallzy (approved):

src/lenten-lentil-curry.md
```diff
@@ -0,0 +1,37 @@
+# Lenten Lentil Curry
+
+Simple green lentil recipe. Suitable for Lent.
```
DGKSK8LIFE, line 3: Capitalization in `Suitable for Lent` seems off hear but correct me if I'm wrong.

xinnna: I apologize for the late response. I think it is correct to capitalize `Lent` here, as it is a specific named period.

DGKSK8LIFE: Ok.

Issue #546, opened 2021-04-03
A tag for recipes with pictures


stnby:
I think we need another tag or something similar that separates all the recipes with pictures from ones that do not contain any pictures. Right now I get an impression that 80% (based on what I clicked) recipes have no pictures. And these ones do not interest me at all and probably some other people as well.

hallzy:
Ya, looks like it is about 72%, which is too high imo.

I don't really like the idea of using the same tagging for pictures though because it is something that could be automated (does the recipe file contain an image link? then this recipe has an image), and it might be useful to have a way to filter out recipes without images but also filter by a tag. Like if you want Japanese recipes that have images.

Something that might be doable in the makefile.

Maybe having the Makefile add a camera emoji or something to the beginning of every recipe link that has an image?

peepopoggers:
I agree with this
Idk the best way to indicate this. Maybe color the bulletpoint?
Maybe add camera emoji after bullet point?
There should be some indication because you might end up eating some cow turd and not know

peepopoggers:
![image](https://user-images.githubusercontent.com/72892531/124035499-874ab580-d9f4-11eb-95b9-fbf183b74d11.png)

What do you guys think about how this looks? I don't think the camera emoji looks that bad. 📷
There could also be a sort by field so you could do sort by type (plain or images), sort by newest/oldest, and sort by name.
or, keep it incredibly basic :|

PR #547, opened 2021-04-03
Added Camera emoji on recipes with images

hallzy:
Added A camera emoji to recipe links that have images (done automatically through the Makefile)

region-20210403-133649

REF: #546

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -89,7 +89,7 @@ blog/index.html: index.md $(ARTICLES) $(TAGFILES) $(addprefix templates/,$(addsu
 		"$$first" || envsubst < templates/article_separator.html; \
 		URL="`printf '%s' "\$$FILE" | sed 's,^$(BLOG_SRC)/\(.*\).md,\1,'`.html" \
 		DATE="$$DATE" \
-		TITLE="`head -n1 "\$$FILE" | sed -e 's/^# //g'`" \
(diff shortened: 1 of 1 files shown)


stnby:
This is brilliant.

LukeSmithxyz:
I like the idea, but I feel like it looks strange.

I don't know if it'd be better or worse after the title.

Another option might be replacing the `<li>` decoration with the emoji, but that might be hard to do with discount and the makefile... Might look stupid anyway.

hallzy:
I think it would be worse personally. All the titles are different lengths so you'd have to visually find the end of the title to see if the emoji is there, whereas as the beginning it is always the same place and is easier to skim through.

I was able to quickly make the cameras the list decorator to see how it looks:

region-20210409-124423

You can do this by adding a class value to the `<li>` in the template, and then in the makefile check if the file has an image, and if it does,  give that <li> a class of 'image' or something... Then in CSS, you can provide the emoji as the decorator for list items that have an image class

I haven't committed any of this yet, just a proof of concept and an image to show how it would look...

The CSS by the way would be this:

```css
li.image {
	list-style-type: none;
}

li.image:before {
	content: '📷';
	margin-left: -23px;
	margin-right: 3px;
}
```
It is possible to actually remove the second rule altogether and put the camera emoji as the list-style-type but the separation between the bullet and the text isn't as nice looking:

region-20210409-124705

Also, apparently setting the list-style-type directly isn't consistently supported.

wpnwbn:
maybe just add tag `photo` for all recipes with photos?

wups:
Why not put the camera emoji behind the title? It's just an additional information after all. Currently it looks too prominent; too important.

PR #557, opened 2021-04-07
Spelling

TTWNO:
Fix a bunch of little spelling errors :)


src/spaghetti-and-meatballs.md
```diff
@@ -46,7 +46,7 @@
 5. Add the tomato paste, stir to blend, and cook for a minute longer.
 6. Add the canned tomatoes, basil, and parsley.
 7. Mash and stir the tomatoes with a potato masher until the sauce starts to thicken, about 5 minutes over medium-high heat.
-8. Reduce heat to low and simmer gently while prepating meatballs, stirring occasionally.
+8. Reduce heat to low and simmer gently while pre-patting meatballs, stirring occasionally.
```
hallzy, line 49:
```suggestion
8. Reduce heat to low and simmer gently while preparing meatballs, stirring occasionally.
```

Could this actually be `preparing` instead? Looks like this step is before the meatball steps

hallzy (changes requested):
Just one fix that may be questionable otherwise looks good.

My only other concern is that I don't think fixing typos is substantive enough to need a contribution to be added to the bottom of each file.

From the `example.md` file from Luke:

> If you add something substantial to an already existing recipe (including an image) you may add your name below with the contribution in parens.

hallzy:
Oh, would you also be able to squash down the commits into one. The single commit could just be 'Spelling Fixes' or something.

TTWNO:
I totally agree about the spelling not being substantial for a contribution. Had autopilot on. Will fix in morning.

DGKSK8LIFE (approved):

TTWNO:
Not sure if I squashed right. Had some issues with the branches. Let me know if it needs more fixing.

hallzy:
Doesn't look like it

If you are using the command line, it would be something like this:

```bash
# Go back to state before that last merge
git reset --hard 24a743311c48f590dd55a6bce5d21a502cd7a6ab

# Start interactive rebase at the beginning of your PR
# This will open a text editor where you select how you want to combine everything
git rebase -i 6446eed^

# In your editor, select 'reword' for the first commit listed, and 'fixup' for every other commit below it.
# Close and save file

# That will open up a new editor where you can give a new commit message... something like "Fixing Spelling"
# Close and save editor

# push how you normally push, but do it with a -f. Squashing commits down rewrites git history so you need the -f in order for the push to work.
git push -f
```

After that push, this PR should update automatically if you pushed to the same branch.

TTWNO:
Oh jeez! It put my contributions back. Sorry about that. Fixing.

hallzy (approved):
Nice, looks good to me

Issue #443, opened 2021-03-23
Little Fourish: Tags automatically prefixed by Emojis


LukeSmithxyz:
I was thinking it'd be nice to prefix mainly nationality tags in the tag cloud with emojis. E.g. `🇮🇳 indian`, `🇫🇷 french`, etc.

Does this sound stupid? I think it would make some things stick out a little more.

This wouldn't be for every tag (that might be information overload), mainly just different cultures, or maybe the fast tags.

Of course I don't want submitters to have to manually include the emojis in the tags, but I'd like to have the Makefile append the flags only on the main page (not on the specific pages). I guess it would also be possible to include the emojis in the titles of the tag pages too.

This could probably all be done with a big case statement, but I'm wondering if other people think it's cringe.

Anyone is welcome to open a PR if you want to go ahead and do it. Just a thought.

Bumbadawg:
It makes sense for flags, aside it'd be fluff. Please consider having those types of tags on a separate line, which would make it stick out more.

hallzy:
I made a PR for this, but I closed it cause I realized there are some things that could use more discussing.

The PR I had was a shell script that had a long string with a emoji and country on each line, and used awk to go through it. I that was over-complicating things obviously, because all we really need is if statements or case statements instead.

I can easily change my script to do that, but there is a different approach I thought of as well. We could use CSS instead:

```css
a[href='@american.html']:before {
    content: '🇺🇸 ';
}
```

Adding something like that for each flag would work, but isn't great of we are wanting to limit the css.

I will leave it for discussion.

PR #564, opened 2021-04-08
Added Flag Emojis

hallzy:
Thought I would post what I have done so far for feedback before wasting my time converting country names to demonyms, because that will take forever.

REF: #443 

Luke mentioned in the above issue the possibility of adding flags to country tags.

I made a shell script that takes a tag name and gets an emoji for  that tag name... Technically it could be any tag if we choose to add emojis for other tags in the future, but it only consists of countries and their flags right now.

The shell script is executed automatically in the Makefile.

I took a list of countries from somewhere so the list is pretty huge and I have not converted country names to their demonyms yet. I have converted all of them that we currently have on this website though so that we can use it as is and see what it looks like.

This is what it looks like on the homepage:

region-20210407-210515

This is what it looks like at the bottom of recipes:

region-20210407-210537

(yes, I recognize that the flag I have for 'english' is not the flag of england, but apparently the flag of england is not that well supported, so the union jack is better than nothing).

If we like how this looks, I can continue doing the conversions. If not, then we can close this.

I also think it is probably overkill to have the huge list of countries that are listed.

Maybe it is enough to only have the countries that we actually use, and add them as we go.


hallzy:
Also... Quebec is not a country... so I don't know if we want to keep it as quebec for specificity, or change it to `canadian`... either way

hallzy:
Realized after making this PR that we could also use css instead of this script... ie

```css
a[href='@american.html']:before {
    content: '🇺🇸 ';
}
```

And just add one of those for every flag we want. And if we go this route, then I'll just close this PR and make a new one

PR #566, opened 2021-04-09
Add Celsius amount to the crouton recipe

almaember1098:
While yes, 350 °F is technically only 176 °C, no Celsius-based oven has such a setting, and the closest one is 180 °C.

### DIFF
--- a/src/croutons.md
+++ b/src/croutons.md
@@ -12,7 +12,7 @@ Croutons are an essential addition to many salads.  They are delicious and easy 
 
 ## Directions
 
-1. Preheat oven to 350 degrees Fahrenheit (roughly 180 degrees Celsius).  Gently cut slices of white bread into squares with a chef's knife.  One to two slices is usually enough for one person.
+1. Preheat oven to 350°F (roughly 180°C).  Gently cut slices of white bread into squares with a chef's knife.  One to two slices is usually enough for one person.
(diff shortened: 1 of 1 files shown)


src/croutons.md
```diff
@@ -12,7 +12,7 @@ Croutons are an essential addition to many salads.  They are delicious and easy
 
 ## Directions
 
-1. Preheat oven to 350 degrees Fahrenheit.  Gently cut slices of white bread into squares with a chef's knife.  One to two slices is usually enough for one person.
+1. Preheat oven to 350 degrees Fahrenheit (roughly 180 degrees Celsius).  Gently cut slices of white bread into squares with a chef's knife.  One to two slices is usually enough for one person.
```
hallzy, line 15:
```suggestion
1. Preheat oven to 350°F (roughly 180°C).  Gently cut slices of white bread into squares with a chef's knife.  One to two slices is usually enough for one person.
```

Thought may as well add the `°` symbol while here.

hallzy (approved):

src/croutons.md
DGKSK8LIFE: Make sure you are using the right encoding (like UTF tags, ASCII, etc), because some browsers may not render the symbol correctly and/or the markdown to html engine may have problems rendering it. @hallzy

DGKSK8LIFE (approved):

PR #574, opened 2021-04-13
Replace em-dashes with -- for HTML purposes

equwal:
Some applications will mess up with these, and since they are also used
in the <title>, &em; won't work either.

So -- is a good solution to this.


hallzy (approved):

DGKSK8LIFE (changes requested):
Would like a review from @LukeSmithxyz

PR #597, opened 2021-04-21
tag-prune

siggines:
This commit removes some tags from existence (across all the recipes they're used in).

- baked https://based.cooking/@baked
- baking https://based.cooking/@baking
- traditional https://based.cooking/@traditional

Are tags which are all both irrelevant and under populated.
Why would there be a baking tag when half of all recipes involve baking? These tags only have one recipe in each, and they are two of the same thing.
Traditional only has three recipes, but many of the other recipes on the site could also be called traditional which are not tagged as such. Silly tag.
Since people keep making tags like these and are going unchecked, perhaps there should be a ban on some tags like 'baked' so that they fail the checks.

### DIFF
--- a/src/australian-snags.md
+++ b/src/australian-snags.md
@@ -29,4 +29,4 @@ especially national events. Sometimes served at hardware chains.
 7. Add onions then tomato sauce on top
 8. Enjoy and serve to everyone
 
-;tags: basic snack traditional australian
+;tags: basic snack australian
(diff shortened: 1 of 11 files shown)


siggines:
Fry, Ham, and Feta, are also questionable. Since fry only has two recipes, Feta and Ham only one. Fry is comprable to 'baking', while ham and feta are sub categories of pork and cheese respectively.
Have not added these to the commit.

LukeSmithxyz:
I might want to add a rule that tags should only be used if there are already three or so dishes for which they could be used.

PR #598, opened 2021-04-22
More simple egg recipes for eggs.md

dther:
I like eggs.

### DIFF
--- a/src/eggs.md
+++ b/src/eggs.md
@@ -53,9 +53,36 @@ The more you scramble the less fluffy the egg will be.
 Scrambled eggs are easy to customize. E.g. start by frying any of: onions, bits
 of ham, mushrooms, chopped tomatoes, etc., before cracking the eggs in.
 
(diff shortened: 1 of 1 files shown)


siggines:
There is already an eggs recipe page, https://based.cooking/eggs , you may want to incoroporate your stuff into that file instead?

siggines:
The 'tips' sections seem a bit verbose to me, making the eggs file a long document. Also most of that advice such as how to whisk properly could be stated in the directions. Just for the sake of being concise maybe remove or shorten those sections?

dther:
> The 'tips' sections seem a bit verbose to me, making the eggs file a long document. Also most of that advice such as how to whisk properly could be stated in the directions. Just for the sake of being concise maybe remove or shorten those sections?

Fair point. Is this better?

siggines (approved):
> Fair point. Is this better?

Lovely, thanks for taking the time.

PR #600, opened 2021-04-22
ANZAC-biscuits, and introduction of a Hamburger tag. *ANZAC should be capitalised on recipe list, as it is an acronym.

siggines:
There are 4 hamburger recipes, so I made a hamburger tag.
And ANZAC-biscuits recipe.


siggines:
^ Not passing check due to violating capitalisation ban. However ANZAC is an acronym, for: Australian and New-Zealand Army Corps.

hallzy:
I'd argue that even though it is an acronym, it should still be lowercase in the filename (as you have with you picture)

siggines:
> I'd argue that even though it is an acronym, it should still be lowercase in the filename (as you have with you picture)

Ok, but people don't see the picture name. As they see the title of the recipe, the capitals show it's an acronym.

PR #601, opened 2021-04-23
Give some love to the scrollbar.

Filius-Patris:
Scrollbars deserve attention too. This commit styles the scrollbar according to the current theme (light/dark) using the `scrollbar-color` attribute ([MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-color)). This attribute is currently only supported by Firefox.

EDIT: There's also support for webkit-based browsers (including chrome-based browsers).

### DIFF
--- a/data/style.css
+++ b/data/style.css
@@ -14,6 +14,7 @@ body {
 }
 ::-webkit-scrollbar-thumb {
 	background-color: gray;
+	border-radius: 5px;
 }
 
 a {
--- a/data/style.css
+++ b/data/style.css
@@ -8,6 +8,13 @@ body {
 	font-family: sans-serif ;
 	scrollbar-color: white #151515;
 }
+::-webkit-scrollbar {
+	width: .66vw;
+	background-color: #151515;
+}
+::-webkit-scrollbar-thumb {
+	background-color: white;
+}
 
 a {
 	color: lightblue ;
@@ -53,6 +60,12 @@ img[alt="XMR Logo"] {
 		background: white ;
 		color: black ;
 		scrollbar-color: black white;
+	}
+	::-webkit-scrollbar {
+		background-color: white;
+	}
+	::-webkit-scrollbar-thumb {
+		background-color: black;
 	}
 	a {
 		color: blue ;
--- a/data/style.css
+++ b/data/style.css
@@ -6,14 +6,14 @@ body {
 	padding: 0 16px ;
 	margin-bottom: 500px ;
 	font-family: sans-serif ;
-	scrollbar-color: white #151515;
+	scrollbar-color: gray #151515;
 }
 ::-webkit-scrollbar {
 	width: .66vw;
 	background-color: #151515;
 }
 ::-webkit-scrollbar-thumb {
-	background-color: white;
+	background-color: gray;
 }
 
 a {
@@ -59,13 +59,13 @@ img[alt="XMR Logo"] {
 	body {
 		background: white ;
 		color: black ;
-		scrollbar-color: black white;
+		scrollbar-color: gray white;
 	}
 	::-webkit-scrollbar {
 		background-color: white;
 	}
 	::-webkit-scrollbar-thumb {
-		background-color: black;
+		background-color: gray;
 	}
 	a {
 		color: blue ;
--- a/data/style.css
+++ b/data/style.css
@@ -6,6 +6,7 @@ body {
 	padding: 0 16px ;
 	margin-bottom: 500px ;
 	font-family: sans-serif ;
+	scrollbar-color: white #151515;
 }
 
 a {
@@ -51,6 +52,7 @@ img[alt="XMR Logo"] {
 	body {
 		background: white ;
 		color: black ;
+		scrollbar-color: black white;
 	}
 	a {
 		color: blue ;


siggines:
Is this the same as the dark scrollbar on github with the dark theme?

DGKSK8LIFE:
> but what I mean is: is it visually the same? If so that is dope and I agree it would look good on the site.

Yes if you could send some screenshots over to show the difference I would appreciate it, thanks.

Filius-Patris:
> Yes if you could send some screenshots over to show the difference I would appreciate it, thanks.

Screenshot 2021-04-23 at 21 59 14
_Before: Light mode_

Screenshot 2021-04-23 at 21 57 30
_Before: Dark mode_

Screenshot 2021-04-23 at 21 58 51
_After: Light mode_

Screenshot 2021-04-23 at 21 58 12
_After: Dark mode_

The colors are the same as text/bg (ie. `white` and `#151515` for dark mode and `black` and `white` for light mode). They might be too strong, I could tune them down a bit to something more greyer.

> Is this the same as the dark scrollbar on github with the dark theme?

You mean this?
Screenshot 2021-04-23 at 22 04 28

Nope, as far as I can tell, that's not explicit styling by CSS, this seems to be recognized by Chrome (or Blink or whatever) and applied. But I might be wrong...

DGKSK8LIFE (approved):

data/style.css
```diff
@@ -6,6 +6,7 @@ body {
 	padding: 0 16px ;
 	margin-bottom: 500px ;
 	font-family: sans-serif ;
+	scrollbar-color: white #151515;
 }
```
siggines, line 11:
```suggestion
::-webkit-scrollbar {
    height: 12px;
    width: 12px;
    background: #151515;
}

::-webkit-scrollbar-thumb {
    background: #000;
    -webkit-box-shadow: 0px 1px 2px rgba(0, 0, 0, 0.75);
}

::-webkit-scrollbar-corner {
    background: #000;
}

```

siggines (changes requested):
Will webkit, I have made a similiar one avaliable in chromium engine browsers.
Please edit colour codes if you think you have superior. I will post screen shot too

data/style.css
siggines: This is actually a bad idea as it is, as zooming into page also bloody makes the scroll bar zoom in, horrid. Bear with me

siggines: Yeah, I would disregard using webkit retrospectively. If only other browsers were as graceful as Firefox.
How does Github make the dark scrollbar?

hallzy: Haven't tried anything, but you can see what github does in the browser devtools:

```css
::-webkit-scrollbar {
    background-color: #202324;
    color: #aba499;
}

::-webkit-scrollbar-corner {
    background-color: #181a1b;
}

::-webkit-scrollbar-thumb {
    background-color: #454a4d;
}
```

Filius-Patris: > Haven't tried anything, but you can see what github does in the browser devtools

Where did you find that? I was looking, but haven't found anything. What element did you select and where did you look?

siggines: > Haven't tried anything, but you can see what github does in the browser devtools:
> 
> ```css
> ::-webkit-scrollbar {
>     background-color: #202324;
>     color: #aba499;
> }
> 
> ::-webkit-scrollbar-corner {
>     background-color: #181a1b;
> }
> 
> ::-webkit-scrollbar-thumb {
>     background-color: #454a4d;
> }
> ```

I have updated my suggested edit to use this, the Github colour codes, thanks.

Filius-Patris: Should I use the same color codes as GitHub light mode too? And how did you guys find these selectors? They don't show up anywhere for me??

siggines: Well since webkit will work for firefox too I assume, the scrollbar-width property is redundant. And IDK how Hallzy found the codes for githubs colour.

siggines, line 11:
```suggestion
::-webkit-scrollbar{
	width:.66vw;
    	background-color: #151515;
    	color: #aba499;
}

::-webkit-scrollbar-corner{
    	background-color: #181a1b;
}

::-webkit-scrollbar-thumb{
    	background-color: #454a4d;
}
```
OK. In this new suggestion, I have made the background scrollbar colour the same colour as the background on based.cooking and I have made it 2/3 of the size to be streamlined aswell.

siggines (changes requested):

data/style.css
siggines: I did not remove the original firefox affecting property made by OP, but this webkit stuff will likely override it on Mozilla.

Filius-Patris: > Well since webkit will work for firefox too I assume, I would say the scrollbar-width property is redundant.

Uhm no, all these `::-webkit-` prefixed pseudoselectors don't affect FF at all. I don't see stylized scrollbars in GH with FF. Anyway, I also don't see stylized scrollbars for GH in Chrome either, I just realized, I just see the dark scrollbar, bc GH has `color-scheme: dark`.

Can anyone of you pls send a screenshot of the scrollbar or the `::-webkit-` selector on GH? I can't find it, nor the colors from there. I would need those for the light mode styling.

siggines: Screenshot from 2021-04-23 22-10-33

Filius-Patris: > I did not remove the original firefox affecting property made by OP, but this webkit stuff will likely override it on Mozilla.

It doesn't: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar#-webkit-scrollbar.

But do you have the colors for light mode too?

siggines: The bg of this test site is the same as that on based.cooking . ( #151515 )

Filius-Patris: Yes, but there are 3 other colors, I have no idea where they show up and also no idea why they are different. Isn't a scrollbar supposed to have two colors? One for the thumb and one for the background (#151515)?

siggines: No but I guess that can be done, erm. If instead of putting the hex code in I used a variable that changes based on the browsers gypsy theme. Can that be done in CSS? I dunno, I'm giving the mantle of that task to someone else I think. 
>It doesn't
Ok, then you may change your colour code to the same as I have it if you want. Or remove your edits in favour of my webkit. It may be good to have it as a redundancy IDK.

siggines: > Yes, but there are 3 other colors, I have no idea where they show up and also no idea why they are different. Isn't a scrollbar supposed to have two colors? One for the thumb and one for the background (#151515)?
That is the THUMB
::-webkit-scrollbar-thumb{
    	background-color: #454a4d;
}

siggines:
Yes, and then the 'scrollbar-color' property may remain as a good redundancy for browsers incapable of webkit but with Mozilla engine. I think there might be a few examples of that like icecat or whatver

data/style.css
```diff
@@ -6,6 +6,14 @@ body {
...
 	margin-bottom: 500px ;
 	font-family: sans-serif ;
+	scrollbar-color: white #151515;
+}
+::-webkit-scrollbar {
+	width: .66vw;
+	background-color: #151515;
+}
+::-webkit-scrollbar-thumb {
+	background-color: white;
```
siggines, line 16:
```suggestion
	background-color: gray;
```
White is incredibly sheer.

siggines (changes requested):

data/style.css
siggines: Screenshot from 2021-04-23 22-30-05
Screenshot from 2021-04-23 22-30-22

Filius-Patris: Ah, yes much more pleasing. Lemme check light mode too first...

data/style.css
```diff
@@ -51,6 +59,13 @@ img[alt="XMR Logo"] {
...
 		background: white ;
 		color: black ;
+		scrollbar-color: gray white;
+	}
+	::-webkit-scrollbar {
+		background-color: white;
+	}
+	::-webkit-scrollbar-thumb {
+		background-color: gray;
 	}
```
siggines, line 69:
```suggestion
	}
	::-webkit-scrollbar-track:vertical:disabled:corner-present {
        margin-bottom: 0px;
    }
```
Might you please put this in also, so that when there is no horizontal scrollbar (only would happen when you're really zoomed in) the corner will not be there.

siggines (changes requested):

data/style.css
```diff
@@ -6,6 +6,14 @@ body {
...
 	font-family: sans-serif ;
+	scrollbar-color: gray #151515;
+}
+::-webkit-scrollbar {
+	width: .66vw;
+	background-color: #151515;
+}
+::-webkit-scrollbar-thumb {
+	background-color: gray;
 }
```
siggines, line 18:
```suggestion
::-webkit-scrollbar-track:vertical:disabled:corner-present {
margin-bottom: 0px;
}
::-webkit-scrollbar-corner{
background-color: #151515;
}
```
It is necessary for OP to style the CORNER for dark mode, cus by deafult it is white.
The top bit of code I added also will actually hide the corner unless there is a horizontal bar (when zoomed in), I prefer this. Luke may disagree and remove that top bit I dunno.

siggines (changes requested):

data/style.css
siggines: Additionaly, if Luke wants a slightly rounded scrollbar:

::-webkit-scrollbar-thumb{      **border-radius: 5px;**  ...

Filius-Patris: IMHO disabling the corner is anything but intuitively comprehensive. I'm more in the favor of styling the corner.

But anyway: Do we need that? The site is perfectly responsive. I checked and found no situation when horizontal scrollbars would appear...

It's a balance of keeping the stylesheet small and comprehensive vs not having to cover that case if it happens in the future. I'm in favor of the former. Do you think it is needed?

Filius-Patris: > Additionaly, for a slightly rounded scrollbar:
Done. Thanks for the suggestion!

siggines: > IMHO disabling the corner is anything but intuitively comprehensive. I'm more in the favor of styling the corner.
> 
> But anyway: Do we need that? The site is perfectly responsive. I checked and found no situation when horizontal scrollbars would appear...
> 
> It's a balance of keeping the stylesheet small and comprehensive vs not having to cover that case if it happens in the future. I'm in favor of the former. Do you think it is needed?

Alright, as long as it looks good from your end (in both light and dark modes).

siggines:
For Lukes reference. Rounded **grey** scroller, with background of #151515, to blend with website background.
Dark-mode-scroll,rounded,slim

linux-leo:
So can we merge this, i feel like that would be an awesome improvement and its been over a month

PR #637, opened 2021-05-25
Add lemonade and taco seasoning recipes

ifrendas:


src/taco-seasoning.md
```diff
@@ -0,0 +1,25 @@
...
+- 1/2 tsp seasoned salt
+- 1/2 tsp paprika
+- 1/2 tsp garlic salt
+- 1/4 tsp cumin
+- 1/4 tsp cayenne pepper
+- 1/4 tsp sugar
+
+## Directions
+
+1. Mix ingredients in a bowl
```
DGKSK8LIFE, line 19: These directions are incomplete.

DGKSK8LIFE (changes requested):

src/taco-seasoning.md
ifrendas: How? Once you mix the ingredients the seasoning is made. It's tagged as basic because it isn't stand alone.

DGKSK8LIFE: Well if that's the only direction, could you elaborate about mixing ingredients into the bowl? Like are you using any specific utensil or?

ifrendas: I changed the wording in the latest commit to hopefully make it more clear.

DGKSK8LIFE (approved):

src/lemonade.md
```diff
@@ -8,9 +8,9 @@ A simple lemonade recipe
 
 ## Ingredients
 
-- 1.75 cups white sugar
+- 1¾ cups white sugar
 - 8 cups water
-- 1.5 cups lemon juice
+- 1½ cups lemon juice
```
DGKSK8LIFE: Imo the previous formatting was better.

DGKSK8LIFE (changes requested):

src/lemonade.md
ifrendas: Ah, my bad. I'll change it back.

DGKSK8LIFE (approved):

PR #643, opened 2021-06-06
Add monero address

Danoloan10:
I added my monero address for donations.


DGKSK8LIFE (approved):
Maybe add the XMR address underneath as a subheader? Just a thought.

Danoloan10:
I'm not sure I understand well I prefer it this way

PR #644, opened 2021-06-10
revise pancake.md

alzheimerspenguin:
The basic pancake recipe needed some work.

Baking Soda needed to be changed to Baking Powder (which is a different ingredient, includes a necessary raising agent for this recipe)

Ratio/serving sizes were added (ie this creates 3 pancakes)

All amounts were changed to easy to use and multiply measurements/ratios
eg. 200g of flour should be measured in cups, you don't pull out a scale to weigh your flour when you make pancakes

Butter was added, nobody should be making pancakes or any type of cake batter without butter, this is standard lindy.

Undefined "oil" was removed - vegetable/seed oils are obviously pure poison and we're not going to use coconut/olive/avocado or any other meme lubricants for a pancake because they will alter (worsen) the taste of the product

Frying instructions were given more detail/specifics.
Batter consistency/viscosity instructions were added

### DIFF
--- a/src/pancake.md
+++ b/src/pancake.md
@@ -2,24 +2,39 @@
 
 ![Pancakes made at home](pix/pancake.webp)
 
-Here is a simple Pancake recipe
+Simple timeless pancake recipe with easy ratios and staple ingredients
 
-## Ingredients
-- 200g wheat flour
-- 200 ml milk
-- 2 Eggs
-- 2 teaspoon sugar
-- 1 teaspoon baking soda
-- a little bit of salt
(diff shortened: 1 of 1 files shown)


DGKSK8LIFE (approved):
Looks good. Please change it to use proper markdown formatting (one line between header and header content), thanks.

alzheimerspenguin:
fixed the formatting, i think.... if i understood correctly.
Also made a slight edit to whisking instructions

DGKSK8LIFE (approved):
Cool. Good job.

PR #654, opened 2021-06-22
Add Medieval Beef Soup

LogSuperMulti:
Adding a modified beef soup recipe from Anonimo Toscano's manuscript.


DGKSK8LIFE (approved):
Looks great. May try this myself.

PR #661, opened 2021-07-04
Fully operational offline website

StrajnarFilip:
This service worker will allow users to have fullly operational website even offline. Every asset they ever load will be cached, allowing the use in remote areas or in cases of bad reception. Strategy used is network first, but there is also a function for cache first approach in case it's more suitable (user would have to manually clear old assets in order for new one to load).


binex-dsk:
man suggested using js

PR #675, opened 2021-07-20
create menemen.md

munzevibirinsan:


Socerest2:
I think we should close this - user deleted their account.

PR #704, opened 2021-10-03
Url-redirect on my recipes and added garlic-sauce & oven-dish-pasta recipes

user18130814200115-2:
- Updated urls to my website (previously twester.tk) now wester.digital
- Added recipe for Garlic Sauce
- Added recipe for oven-dish-pasta


LukeSmithxyz:
These changed were added to the new site.

PR #729, opened 2022-01-17
added chicken tikka masala

aaronbjohnson:

### DIFF
--- a/src/chicken-tikka-masala.md
+++ b/src/chicken-tikka-masala.md
@@ -1,0 +1,36 @@
+# Chicken Tikka Masala
+
+- ⏲️ Prep time: 25 min
+- 🍳 Cook time: 4.25 hours
+- 🍽️ Servings: 8
+
+## Ingredients
+
+- 1 can (29 ounces) tomato puree
+- 1-1/2 cups plain yogurt
+- 1/2 large onion, finely chopped
+- 2 tablespoons olive oil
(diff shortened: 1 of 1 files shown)


ThamognyaKodi (commented):
Great recipe. Could you convert min to an hour or vice versa for all of the units, I think a standardized time unit is better.

hmccarty:
> Great recipe. Could you convert min to an hour or vice versa for all of the units, I think a standardized time unit is better.

Varying denominations of time is the standard, can't think of a good reason to write 240 minutes instead of 4 hours. 

Good recipe though, should merge PR.

aaronbjohnson:
> > Great recipe. Could you convert min to an hour or vice versa for all of the units, I think a standardized time unit is better.
> 
> Varying denominations of time is the standard, can't think of a good reason to write 240 minutes instead of 4 hours.
> 
> Good recipe though, should merge PR.

That's what I thought as well. I feel like most people would prefer 30 minutes to .5 hours etc.

aaronbjohnson:
> > Great recipe. Could you convert min to an hour or vice versa for all of the units, I think a standardized time unit is better.
> 
> Varying denominations of time is the standard, can't think of a good reason to write 240 minutes instead of 4 hours.
> 
> Good recipe though, should merge PR.

Actually I just realized that it shouldn't give a total cooking time of 4.25 hours...that is a bit weird. It should be just 4 hours cooking. I'll make the change.

PR #731, opened 2022-01-23
Added scrambled-eggs.md: A Scrambled eggs recipe

ThamognyaKodi:
This is an scrambled eggs recipe.


binex-dsk:
90% cooked? Severely overcooking. Cook the one side until you can easily move it around without it sticking, and cook the other side for 30 seconds at most.

PR #756, opened 2022-04-17
Add Overnight Oats

ctrl-alt-w:

### DIFF
--- a/content/overnight-oats.md
+++ b/content/overnight-oats.md
@@ -1,0 +1,26 @@
+---
+title: Overnight Oats
+tags: ['breakfast', 'easy', 'oats']
+date: 2022-04-17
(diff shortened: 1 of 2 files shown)


peepopoggers:
I feel like mental outlaw's recipe would be better, check it out

content/overnight-oats.md
```diff
@@ -0,0 +1,26 @@
+---
+title: Overnight Oats
+tags: ['breakfast', 'easy', 'oats']
+date: 2022-04-17
+author: "Ctrl-Alt-W"
+---
+
+- ⏲️ Prep time: 5 min
+- 🍳 Cook time: 0 min
```
Newbytee, line 9: Is the cook time really 0 minutes if it needs to sit overnight?

PR #768, opened 2022-05-19
Create recipe Apple crumble with speculaas

kryzar:
Add recipe `apple-crumble-speculaas.md`.


LukeSmithxyz:
That photo is obscenely massive (1.56 MiB). Remove it or reduce its size.

kryzar:
The new size is 270k. Tell me if you want more compression or anything else.

LukeSmithxyz:
The image is still 4000x3000 and would still be the largest image on the site by far (including others that really should be shrunk). Note that the CSS-stipulated maximum for image height is 400px. It's fine for images to natively be larger, but not that much larger.

kryzar:
Done. We should maybe add this constraint (height <= 400px) to the README. I can make a pull request if you want.

PR #769, opened 2022-05-22
Add steak tartare recipe

wdunkelheit:


wdunkelheit:
rebased corrections into single commit to minimise number of commits for single entry, force pushed on my branch to overwrite the multiple commits

PR #772, opened 2022-05-26
Filter Articles with a Javascript Search

LukeSmithxyz:
This PR is from @omar2205 and #533 updated for the Hugo site.

The number of articles is becoming unwieldy and it'd be nice to filter them quickly. This is theoretically what Javascript *should* be doing.

I almost want to merge this as is, but there are a few improvements I want, but I don't know how to best do them in JS.

1. Right now, the user has to type in the search bar and press enter to filter articles. I would like to remove the enter step and have them filter real-time.
2. I'd like to use JS to display the filtered articles larger, i.e. change CSS. I don't know how to do this.
3. Note that I've removed the tag/category list with this PR, but I'd like suggestions of what to do here as well. Categories still display at the end of each recipe, but I feel like there should be a category list on the mainpage. Having it and the search seems blinding.

If anyone can aid here, please say so. I think this feature will make this site significantly more normie-friendly and more of non-LARP.

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -52,9 +52,11 @@ document.addEventListener('DOMContentLoaded', () => {
 })
 </script>
 
-## All Recipes
-
 {{< artlist >}}
+
+## Or Browse by Category...
+
+{{< tagcloud >}}
 
 ## About this site
 
--- a/content/_index.md
+++ b/content/_index.md
@@ -20,7 +20,7 @@ document.addEventListener('DOMContentLoaded', () => {
   search.addEventListener('change', e => {
     // grab search input value
     const searchText = e.target.value.toLowerCase()
-    // for each receipe hide all but matched
+    // for each recipe hide all but matched
     rec.forEach(el => {
       const recipeName = el.innerText.toLowerCase()
       const isMatch = recipeName.includes(searchText)
--- a/content/_index.md
+++ b/content/_index.md
@@ -48,7 +48,6 @@ document.addEventListener('DOMContentLoaded', () => {
       el.classList.remove('matched-recipe');
     })
 
-    ul.classList.remove('is-filtered');
     artlist.classList.remove('list-searched') ;
   })
 })
--- a/content/_index.md
+++ b/content/_index.md
@@ -17,7 +17,7 @@ document.addEventListener('DOMContentLoaded', () => {
   const search = document.querySelector('#search')
   const clearSearch = document.querySelector('.clear-search')
 
-  search.addEventListener('change', e => {
+  search.addEventListener('input', e => {
     // grab search input value
     const searchText = e.target.value.toLowerCase()
     // for each recipe hide all but matched
--- a/content/_index.md
+++ b/content/_index.md
@@ -21,7 +21,11 @@ document.addEventListener('DOMContentLoaded', () => {
   search.addEventListener('input', e => {
     // grab search input value
     const searchText = e.target.value.toLowerCase()
+
+    const hasFilter = searchText.length > 0;
+
     // for each recipe hide all but matched
+    let matchCount = 0;
     rec.forEach(el => {
       const recipeName = el.innerText.toLowerCase()
       const isMatch = recipeName.includes(searchText)
--- a/content/_index.md
+++ b/content/_index.md
@@ -2,9 +2,39 @@
 title: "🍲 Based Cooking 🍳"
 ---
 
-## Categories
+## What do you want to cook?
 
-{{< tagcloud >}}
+<div class="search">
+  <input type="text" id="search" placeholder="Search...">
+  <button class="clear-search">
+    <svg xmlns="http://www.w3.org/2000/svg" class="ionicon" viewBox="0 0 512 512"><title>Backspace</title><path d="M135.19 390.14a28.79 28.79 0 0021.68 9.86h246.26A29 29 0 00432 371.13V140.87A29 29 0 00403.13 112H156.87a28.84 28.84 0 00-21.67 9.84v0L46.33 256l88.86 134.11z" fill="none" stroke="currentColor" stroke-linejoin="round" stroke-width="32"></path><path fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="32" d="M336.67 192.33L206.66 322.34M336.67 322.34L206.66 192.33M336.67 192.33L206.66 322.34M336.67 322.34L206.66 192.33"></path></svg>
+  </button>
+</div>
+
+<script>
+document.addEventListener('DOMContentLoaded', () => {
+  const rec = document.querySelectorAll('#artlist li')
+  const search = document.querySelector('#search')
+  const clearSearch = document.querySelector('.clear-search')
+
+  search.addEventListener('change', e => {
+    // grab search input value
+    const searchText = e.target.value.toLowerCase()
+    // for each receipe hide all but matched
+    rec.forEach(el => {
+      const recipeName = el.innerText.toLowerCase()
+      const isMatch = recipeName.includes(searchText)
+
+      el.hidden = !isMatch
+    })
+  })
+
+  clearSearch.addEventListener('click', e => {
+    search.value = ''
+    rec.forEach(el => el.hidden = false)
+  })
+})
+</script>
 
 ## All Recipes
 
--- a/content/_index.md
+++ b/content/_index.md
@@ -38,7 +38,7 @@ document.addEventListener('DOMContentLoaded', () => {
       }
     })
 
-    ul.classList.toggle('is-filtered', matchCount > 0);
+    artlist.classList.toggle('list-searched', matchCount > 0);
   })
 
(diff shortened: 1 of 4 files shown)


content/_index.md
```diff
@@ -2,9 +2,39 @@
...
+  </button>
+</div>
+
+<script>
+document.addEventListener('DOMContentLoaded', () => {
+  const rec = document.querySelectorAll('#artlist li')
+  const search = document.querySelector('#search')
+  const clearSearch = document.querySelector('.clear-search')
+
+  search.addEventListener('change', e => {
```
hallzy, line 20:
```suggestion
  search.addEventListener('input', e => {
```

Changing this event to 'input' will cause this event to fire every time the input changes, no need for pressing enter anymore

content/_index.md
```diff
@@ -2,9 +2,39 @@
...
+    // grab search input value
+    const searchText = e.target.value.toLowerCase()
+    // for each receipe hide all but matched
```
hallzy, line 23:
```suggestion
    // for each recipe hide all but matched
```

Random spelling fix

content/_index.md
```diff
@@ -2,9 +2,39 @@
...
+    rec.forEach(el => {
+      const recipeName = el.innerText.toLowerCase()
+      const isMatch = recipeName.includes(searchText)
+
+      el.hidden = !isMatch
```
hallzy, line 28:
```suggestion
      el.classList.toggle('matched-recipe', isMatch && searchText.length !== 0);
```

Adds a CSS class `matched-recipe` to the recipes that match the search text, but the class removed from every recipe if the search box is empty

static/style.css
```diff
@@ -100,3 +113,36 @@ img[alt="XMR Logo"] {
...
+  top: 5px;
+  height: 30px;
+  width: 30px;
+  color: #888;
+  cursor: pointer;
+  transition: color 180ms ease-in-out;
+}
+button.clear-search:hover {
+  color: #eee;
+}
```
hallzy, line 148:
```suggestion
}

.matched-recipe {
  font-size: 1.5em;
}
```

Add CSS for the matched items. Can be whatever.

content/_index.md
hallzy: Doing this change exactly as given would not remove the non-matching elements anymore though.

hallzy (commented):
I've suggested some changes that should get you the behaviour you're looking for.

content/_index.md
LukeSmithxyz: Is there anything amiss with just not removing the original line? Seems to work for what I want fine like that.

This actually brings up another issue though. The article list `ul` has responsive CSS that adjusts the number of columns depending on the window size.

Could the `ul`'s CSS too?

hallzy:
```suggestion
      el.hidden = !isMatch
      el.classList.toggle('matched-recipe', isMatch && hasFilter);
```

Nope, you can definitely leave that line there just like this.

We could have this code modify the `ul`'s CSS as well, though it probably wouldn't make sense here, since this is inside a loop, and we really only need to set a class for the `ul` once.

I'll redo this suggestion.

content/_index.md
```diff
@@ -2,9 +2,39 @@
...
+<div class="search">
+  <input type="text" id="search" placeholder="Search...">
+  <button class="clear-search">
+    <svg xmlns="http://www.w3.org/2000/svg" class="ionicon" viewBox="0 0 512 512"><title>Backspace</title><path d="M135.19 390.14a28.79 28.79 0 0021.68 9.86h246.26A29 29 0 00432 371.13V140.87A29 29 0 00403.13 112H156.87a28.84 28.84 0 00-21.67 9.84v0L46.33 256l88.86 134.11z" fill="none" stroke="currentColor" stroke-linejoin="round" stroke-width="32"></path><path fill="none" stroke="currentColor" stroke-linecap="round" stroke-linejoin="round" stroke-width="32" d="M336.67 192.33L206.66 322.34M336.67 322.34L206.66 192.33M336.67 192.33L206.66 322.34M336.67 322.34L206.66 192.33"></path></svg>
+  </button>
+</div>
+
+<script>
+document.addEventListener('DOMContentLoaded', () => {
+  const rec = document.querySelectorAll('#artlist li')
```
hallzy, line 16:
```suggestion
  const ul = document.querySelector('#artlist')
  const rec = document.querySelectorAll('#artlist li')
```

content/_index.md
```diff
@@ -2,9 +2,39 @@
...
+  const search = document.querySelector('#search')
+  const clearSearch = document.querySelector('.clear-search')
+
+  search.addEventListener('input', e => {
+    // grab search input value
+    const searchText = e.target.value.toLowerCase()
+    // for each recipe hide all but matched
+    rec.forEach(el => {
```
hallzy, line 25:
```suggestion
    // grab search input value
    const searchText = e.target.value.toLowerCase()

    const hasFilter = searchText.length > 0;

    // for each recipe hide all but matched
    let matchCount = 0;
    rec.forEach(el => {
```

content/_index.md
```diff
@@ -2,9 +2,39 @@
...
+      const recipeName = el.innerText.toLowerCase()
+      const isMatch = recipeName.includes(searchText)
+
+      el.hidden = !isMatch
+    })
+  })
```
hallzy, line 33:
```suggestion
      if (hasFilter && isMatch) {
        matchCount++;
      }
    })

    ul.classList.toggle('is-filtered', matchCount > 0);
  })
```

content/_index.md
LukeSmithxyz: I'll add that if you press the delete button, it does not remove the highlighted item's class alignment.

content/_index.md
hallzy: Adds `hasFilter`, just for a bit of readability, and `matchCount` to count how many recipes match the search.

content/_index.md
hallzy: New Selector for the `ul`.

content/_index.md
```diff
@@ -2,9 +2,39 @@
...
+
+  clearSearch.addEventListener('click', e => {
+    search.value = ''
+    rec.forEach(el => el.hidden = false)
```
hallzy, line 37:
```suggestion
    rec.forEach(el => {
      el.hidden = false
      el.classList.remove('matched-recipe');
    })

    ul.classList.remove('is-filtered');
```

hallzy: This removes the classes that were added when the clear button is clicked.

content/_index.md
hallzy: Tracks how many times a match was found, and if at least 1 match was found with a search, a class is applied to the `ul`.

content/_index.md
LukeSmithxyz: I commit while you were writing this. Is my solution for getting the article list acceptable? Not sure what's computationally efficient or not in JS lol

content/_index.md
LukeSmithxyz: What's the reason for this exactly? Is there any harm in not having this check?

LukeSmithxyz: Nevermind, I'm dumb. Get it now.

hallzy: So, the way you have it now, the 'list-searched' class will be added to artlist every time the loop runs, which could be wasteful depending on how JS internally does this.

The main reason I did it this way is that if you use the count then if there are no matches then it also removes the class. I guess whether or not that is what you want is something different. If you don't, then isMatch can be removed.

I'd still keep the `hasFilter` check though, because otherwise if the user backspaces the clear the search instead of clicking the clear button, the class won't get removed.

content/_index.md
hallzy: The way you've done it is fine. I'm not sure which is better but If anything it's probably yours anyways.

LukeSmithxyz:
Thanks for the help. It's working how I want now. Are there any other inefficiencies I may've snuck into it?

If not, I might add this pretty soon with some other superficial changes.

hallzy:
No problem. Looks good to me. 👍

LukeSmithxyz:
Great. Thanks again. I'll go ahead an merge it!

content/_index.md
```diff
@@ -2,9 +2,55 @@
...
+  search.addEventListener('input', e => {
+    // grab search input value
+    const searchText = e.target.value.toLowerCase()
+
+    const hasFilter = searchText.length > 0;
+
+    // for each recipe hide all but matched
+    let matchCount = 0;
+    rec.forEach(el => {
+      const recipeName = el.innerText.toLowerCase()
```
hallzy, line 30:
```suggestion
      const recipeName = el.textContent.toLowerCase()
```

Could make this change too... it makes the search way faster.

LukeSmithxyz: K. I'll add it in the next commit.

hallzy:
ah, sry... just missed you.

innerText takes like 1ms to run, which adds up when you have over 200 recipes... textContent is much faster

PR #774, opened 2022-05-31
Add vim-like key-bind to search input

wlwww16:
This simple commit adds an EventListener for when a key is pressed and, if that key is the forward-slash ("/") key, the search input is focused and the user can start typing.

If the search input is already focused, nothing new happens as there is a return statement.

This binding takes inspiration from vim.


LukeSmithxyz:
Binding keys is a bridge I'm just not willing to cross. It has potential to interfere with plugins (e.g. preexisting vim-like plugins) and opens the door for binding even more.

If people want to jump to the search from the top, they can press tab like on any other site.

PR #801, opened 2022-07-01
clean up js

alpheratz0:
* clean up javascript
* set element id instead of class on the `clear-search` button

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -7,7 +7,7 @@ description: 'The fast-loading recipe site with cooking only and no ads.'
 
 <div class="search js-only">
   <input type="text" id="search" placeholder="Search...">
-  <button class="clear-search">
+  <button id="clear-search">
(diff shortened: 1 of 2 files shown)


content/_index.md
```diff
@@ -7,53 +7,49 @@ description: 'The fast-loading recipe site with cooking only and no ads.'
...
-      el.classList.toggle('matched-recipe', isMatch && searchText.length !== 0);
-      if (hasFilter && isMatch) {
-        matchCount++;
-      }
-    })
+    recipes.forEach(recipe => {
+      const recipeName = recipe.textContent.toLowerCase();
+      const isMatch = recipeName.includes(searchText);
 
-    artlist.classList.toggle('list-searched', matchCount > 0);
```
AntonLee, line 46: Note: this line moved to line 33 and no longer accounts for zero matched recipes

alpheratz0: I don't see the problem with that. If there are 0 matches you won't see any visual changes whether or not `list-searched` is toggled on. Also I think that having `list-searched` active while searching/filtering is more appropiate.

PR #804, opened 2022-07-01
convert tags to lowercase

alpheratz0:


LukeSmithxyz:
I see. Although really I need to make the Hugo portion case-non-sensitive.

PR #820, opened 2022-07-08
Add recipes for apple chicken and farci tomatoes

marhu98:
What the title says.


content/apple-chicken.md
```diff
@@ -0,0 +1,37 @@
+---
+title: Apple Chicken
+tags: ['spanish', 'chicken', 'oven']
+date: 2022-06-22
```
alpheratz0, line 4: here

content/farci-tomatoes.md
```diff
@@ -0,0 +1,36 @@
+---
+title: Farcies tomatoes
+tags: ['french', 'tomatoes', 'oven']
+date: 2022-06-22
```
alpheratz0, line 4: and here, you should put 2022-07-08 otherwise it will not show at the top of the page

marhu98: Ah thanks! I wrote it sometime ago and forgot to update the date. Sorry!

PR #845, opened 2022-07-28
Very simple pancakes

scohenjanes5:


BeBel42:
Step 4 is written as step 3, so we have two step 3...

scohenjanes5:
My bad. Just updated it.

PR #856, opened 2022-08-08
Risengrød

vilhelmbergsoe:
Risengrød is part of the Nordic Christmas tradition as a dinner dish. Rice boiled with milk and a little sugar to form porridge. It is served warm with butter and cinnamon mixed with sugar.


larshb:
I think it's called [Rice Pudding](https://www.wikiwand.com/en/Rice_pudding) in english. Alternatively "Traditional Norwegian Rice Pudding".

PR #870, opened 2022-09-01
move script from _index.md to static dir

shaharbest:
image


LukeSmithxyz:
If this were loaded by other pages, it would make sense, but I'd rather avoid the extra file call here. Sure, it's better in principle though.

PR #886, opened 2022-09-26
Add fairy bread recipe

lukedev820:
I noticed there was only one Australian recipe. I decided to contribute a classic Aussie snack.


beomusxyz:
i second this motion

PR #907, opened 2022-12-28
Add reverse seared steak

calebstein1:


hdntc:
sounds yummy when can we get this on the site?

PR #919, opened 2023-02-20
Add a random recipe button

M4xi1m3:
This PR aims to add a random recipe button with as little Javascript as possible. I had to edit the HTML and CSS a bit to fit the new button next to the search bar, but the final result looks good in my opinion.

### DIFF
--- a/content/_index.md
+++ b/content/_index.md
@@ -6,10 +6,15 @@ layout: single
 
 ## What do you want to cook?
 
-<div class="search js-only">
-  <input type="text" id="search" placeholder="Search ALL Recipes...">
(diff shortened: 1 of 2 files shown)


janurskremer:
Great idea!

LukeSmithxyz:
I'll keep this open. I noticed that publicdomainrecipes.com has this. But I'm not sure about adding it. Realistically it doesn't seem to have many use cases...

I have thought about a random element to pick the recipes to display on the mainpage. After all, there's little point in what we have now (showing first the most recent). Perhaps having the first of a random sort of them would be better.

PR #924, opened 2023-03-24
Mead

lpreston-EE:
I wrote a tutorial for brewing mead, a simple honey wine.


Reisyukaku:
As someone that used to brew cider, it may be worth mentioning that the containers should be sanitized (i used one step), also what kind of yeast? Different yeast has different alcohol tolerance. etc

PR #981, opened 2024-11-11
Modern Borscht recipe

skunkee-hub:
There's a borscht recipe already, but it's not based enough.


LukeSmithxyz:
Added via merge from publicdomainrecipes.com

PR #993, opened 2025-10-09
Create clam-chowder

chaotimliang:
Basic clam chowder base.


chaotimliang:
Left salt in recipe, will remove.

